PR #453, opened 2022-08-11
test: integration test SWB env with DS

SanketD92:
Issue #, if available:
GALI-1608

Description of changes:
Adding integration test for creating environment with dataset mounted. Test checks for: 
- DS creation
- Access point creation
- Environment provisioning with DS mounted
- DS link to access point and Env link to access point
- Post-test cleanup logic included

Checklist:


solutions/swb-reference/integration-tests/support/complex/dataSetHelper.ts
```diff
@@ -0,0 +1,52 @@
+import { AccessPoint } from '@aws-sdk/client-s3-control';
+import { AwsService } from "@aws/workbench-core-base";
+import _ from "lodash";
+import ClientSession from "../clientSession";
+
+export class DataSetHelper {
```
nguyen102, line 6: Should we call this `DatasetHelper` to keep the naming consistent? I notice in other class `Set` is lowercase

solutions/swb-reference/integration-tests/tests/isolated/datasets/create.test.ts
```diff
@@ -0,0 +1,116 @@
...
+          new HttpError(400, {
+            statusCode: 400,
+            error: 'Bad Request',
+            message: "requires property 'datasetName'"
+          })
+        );
+      }
+    });
+  });
+  test('path', async () => {
```
nguyen102, line 53: Should the following tests be nested under the describe of `missing parameters`?

solutions/swb-reference/integration-tests/tests/multiStep/dataset.test.ts
```diff
@@ -0,0 +1,86 @@
...
+describe('multiStep dataset integration test', () => {
+  const setup: Setup = new Setup();
+  let adminSession: ClientSession;
+  let datasetsBucketName: string;
+
+  beforeAll(async () => {
+    adminSession = await setup.getDefaultAdminSession();
+    const aws = setup.getMainAwsClient();
+    const { S3_DATASETS_BUCKET_ARN_OUTPUT_KEY } = getConstants();
+    const {
```
nguyen102, line 21: When the API deploy happens, a `.js` file is generated in `swb-reference/src/config`. That file has the CFN output of the main stack. Should we read values from that file instead of pulling it from CF? Technically that file should always be present and we can pull values from it.However, that does couple the API test with the API deployment.

Thoughts @SanketD92 @FernandoAranda

SanketD92: I thought about decoupling them too, but I guess the integration test framework will eventually depend on the API deployment configs and outputs anyways (like we depend on constants that are linked to the stage config file here). 

You bring up a good point about using the `.js` file, saves a CFN SDK call. I'll make that change.

However this becomes a requirement - running cdk-deploy from the machine before running integration tests. I'll write that note in the test example file.

SanketD92: Merging both into integration test config settings actually works best, since we don't need to re-declare some attributes anymore.

SanketD92: Will be covering this in a separate PR as discussed.

solutions/swb-reference/jest.config.js
```diff
@@ -24,8 +24,18 @@ async function init() {
     fs.readFileSync(join(__dirname, `integration-tests/config/${stage}.yaml`), 'utf8')
   );
 
+  const apiStackOutputs = JSON.parse(
+    fs.readFileSync(join(__dirname, `src/config/${process.env.STAGE}.json`), 'utf8')
+  );
+  const apiStackName = Object.entries(apiStackOutputs).map(([key, value]) => key)[0]; //output has a format { stackname: {...props} }
+  const outputs = apiStackOutputs[apiStackName];
+
+  const mainAccountId = outputs.dynamoDBTableOutput.split(':')[4];
```
FernandoAranda, line 33: Can we return the main Account Id as parameter as well ?, I am not sure but I think we actually have the account number [here](https://github.com/awslabs/solution-spark-on-aws/blob/main/solutions/swb-reference/src/SWBStack.ts#L500:~:text=resources%3A%20%5B%60arn%3Aaws%3Acatalog%3A%24%7Bthis.region%7D%3A%24%7Bthis.account%7D%3Aportfolio/*%60%5D)  and we can just set it as output

solutions/swb-reference/scripts/generateCognitoTokens.js
```diff
@@ -20,8 +20,14 @@ const config = yaml.load(
   fs.readFileSync(join(__dirname, `../integration-tests/config/${process.env.STAGE}.yaml`), 'utf8') // nosemgrep
 );
 
-const clientId = config.clientId;
-const userPoolId = config.userPoolId;
+const apiStackOutputs = JSON.parse(
+  fs.readFileSync(join(__dirname, `../src/config/${process.env.STAGE}.json`), 'utf8') // nosemgrep
+);
```
FernandoAranda, line 25: could we try catch and show a friendly error in case they want to run this before setting up the API

solutions/swb-reference/jest.config.js
```diff
@@ -24,8 +24,18 @@ async function init() {
     fs.readFileSync(join(__dirname, `integration-tests/config/${stage}.yaml`), 'utf8')
   );
 
+  const apiStackOutputs = JSON.parse(
```
FernandoAranda, line 27: could we try catch and show a friendly error in case they don't have the json file

solutions/swb-reference/jest.config.js
SanketD92: I'm not sure if we need to export it for integration tests since we can derive it from one of many other stack outputs, like we're doing here.

FernandoAranda (approved):

PR #479, opened 2022-08-25
chore: separate UI common packages

SanketD92:
Issue #, if available:
GALI-1733

Description of changes:
Currently all SWB UI components live in the `solutions/swb-ui` package. There are common components within these that are consumed by most of our UI pages found in:

- `solutions/swb-ui/src/components`
- `solutions/swb-ui/src/common`
- `solutions/swb-ui/src/context`

Eventually we want SWB UI to be comprised of individually UI libraries imported from workbench-core which would need the above common components to be built. Therefore moving these common components to their own `workbench-core/common-ui` package is needed.

Checklist:


workbench-core/common-ui/.eslintignore
```diff
@@ -0,0 +1,6 @@
+.eslintrc.js
+node_modules
+lib
+temp
+src/common/
```
insignias, line 5: @SanketD92 this needs to be fixed !

workbench-core/common-ui/.eslintignore
```diff
@@ -0,0 +1,6 @@
...
+src/index.tsx
```
insignias, line 6: This needs to be fixed as well !

workbench-core/common-ui/src/common/splitPanel.tsx
```diff
@@ -27,8 +27,6 @@ export const getPanelContent = (items: any, itemType: string): any => {
     };
   }
 
-  const item = items[0];
```
FernandoAranda, line 30: Non related to PR, since we are not showing the cost panel this functions are not being used,
but this functions are very specific to that cost panel, is this worth moving to common-ui?
any thoughts on this @SanketD92  @nguyen102 ?

workbench-core/common-ui/src/common/utils.ts
```diff
@@ -1,11 +1,13 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
FernandoAranda, line 1: good idea, thats gonna save some lines in utils

workbench-core/common-ui/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Getting Started with Create React App
+
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+
+## Available Scripts
+
+In the project directory, you can run:
+
+### `npm start`
```
FernandoAranda, line 14: would this command work when there is no ui to show?

PR #489, opened 2022-09-14
chore: separate env UI

SanketD92:
Issue #, if available:
GALI-1710

Description of changes:
Separating out SWB UI components into their respective packages. Since environments-ui package depends on listing components from multiple other packages (like projects and datasets) for listing their resources, we create a separate UI package called `accounts-ui`.

Checklist:


workbench-core/projects-ui/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@aws/workbench-core-projects-ui",
```
nguyen102, line 2: Thoughts on naming this `workbench-core-accounts-ui`? 

I'm thinking we can put the following API components into one package.

`accounts, cost indexes, projects`

Then we can do the same thing with UI components.

If we don't group them  together, we'll end up with a package for each resource entity, which is probably too many packages.

workbench-core/environments-ui/src/pages/environments/new.tsx
```diff
@@ -0,0 +1,337 @@
...
+    {
+      field: 'envTypeConfigId',
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      condition: (a: any) => !!a,
+      message: 'Configuration is Required'
+    },
+    {
+      field: 'description',
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      condition: (a: any) => !a || a.length <= 500,
```
aws-tyler, line 107: Fun fact: I _think_ if you did this as an `&&` and `a` was false, the interpreter would stop checking the rest of the argument after the first false condition, allegedly saving you some time. That is obviously an extreme over optimization for this situation, but  I just always thought that was neat optimization.

nguyen102 (approved):

insignias:
@SanketD92 when you create a new package, please update the [CODEOWNERS](https://github.com/awslabs/solution-spark-on-aws/blob/chore-separate-env-UI/.github/CODEOWNERS) with the responsible group

nguyen102 (approved):

nguyen102 (approved):

DaveMadsen (approved):

workbench-core/environments-ui/src/pages/environments/new.tsx
```diff
@@ -0,0 +1,337 @@
...
+      text: 'Service Workbench',
+      href: '/'
+    },
+    {
+      text: 'Workspaces',
+      href: '/environments'
+    },
+    {
+      text: 'Create Workspace',
+      href: '/environments/new'
```
bhudnell, line 63: If this is going to be a common component, these fields should be configurable

workbench-core/environments-ui/src/pages/environments/new.tsx
```diff
@@ -0,0 +1,337 @@
...
+    }
+  ];
+  const validationRules = [
```
bhudnell, line 66: At the very least, the messages should be configurable to not tie to the usage of "Workspace"

workbench-core/environments-ui/src/pages/environments/new.tsx
```diff
@@ -0,0 +1,337 @@
...
+          [`${field}Error`]: rule.message
+        }));
+        return false;
+      }
+    }
+    setFormErrors((prevState: CreateEnvironmentFormValidation) => ({ ...prevState, [`${field}Error`]: '' }));
+    return true;
+  };
+  const onSelectEnvType = async (selection: EnvTypeItem[]): Promise<void> => {
+    const selected = (selection && selection[0]) || undefined;
```
bhudnell, line 133:
```suggestion
    const selected = selection[0];
```

bhudnell: since selection must be an array, this would work since `selection[0]` would return undefined for an empty array

workbench-core/environments-ui/src/pages/environments/new.tsx
```diff
@@ -0,0 +1,337 @@
...
+      envTypeId: selected?.id,
+      envTypeConfigId: undefined,
+      envType: selected?.type,
+      datasetIds: []
+    });
+    validateField('envType', selected?.id);
+    validateField('envTypeConfigId', undefined);
+  };
+  const onSelectEnvTypeConfig = (selection: EnvTypeConfigItem[]): void => {
+    const selected = (selection && selection[0]) || undefined;
```
bhudnell, line 146:
```suggestion
    const selected = selection[0];
```

bhudnell (commented):
If these components are going to be used outside of SWB, a lot of the UI text should probably be configurable to not tie users to the SWB naming conventions, ie workspace, environments, etc

workbench-core/environments-ui/src/pages/environments/new.tsx
SanketD92: This doesn't work

workbench-core/environments-ui/src/pages/environments/new.tsx
SanketD92: This doesn't work

workbench-core/environments-ui/src/pages/environments/new.tsx
SanketD92: Another package would be required to handle customizable strings, which is out of scope of this PR. We can work on that once more than one solution depends on this package.

workbench-core/environments-ui/src/pages/environments/new.tsx
SanketD92: Same comment as above

PR #504, opened 2022-10-03
chore: move accounts logic in new package

SanketD92:
Issue #, if available:
GALI-1775

Description of changes:
Moving business logic related to accounts to its new package.

Checklist:


workbench-core/accounts/src/constants/accountResourceTypeToKey.ts
```diff
@@ -0,0 +1,16 @@
+/*
```
nguyen102, line 1: I think we might have to put this in `workbench-core-base`. Doesn't `workbench-core-environment` require this value?

Also should we move this [file](https://github.com/awslabs/solution-spark-on-aws/blob/b045c4c0d21724d13ae7e6e92c6dda000e6e7dc6/workbench-core/environments/src/constants/environmentResourceTypeToKey.ts) to base? 

Since we're referencing the same constant for `PROJ`, we probably want to define it in one place.

SanketD92: Makes sense. I'll create a file in base package to store all resource constants globally.

nguyen102 (approved):

bhudnell (commented):
Looks like there is a CodeQL alert. If it is not a concern, let me know and I'll approve

bhudnell (approved):

PR #510, opened 2022-10-05
chore: Update Environment Model

aws-tyler:
Issue #, if available:
GALI-1789
GALI-1790

Description of changes:

* Removed datasetIds from Environment object.
* Changed type on Environment to be the sk value of the EnvironmentTypeConfig.
* Updated tests.

Checklist:

* [X] Have you written new tests for your core changes, as applicable?


solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
```diff
@@ -129,9 +128,8 @@ describe('SagemakerNotebookEnvironmentLifecycleService', () => {
     expect(response).toEqual({ ...environment, status: 'PENDING' });
   });
 
-  test('Launch should return mocked id when mounting datasets', async () => {
+  test.skip('Launch should return mocked id when mounting datasets', async () => {
```
aws-tyler, line 131: @nguyen102 I'm pretty sure this test (and possibly the one above it) isn't doing what we think it is, so I marked it as `skip` but I could also remove it completely. We are mocking `environment` which gets passed into `sm.launch`, is never modified, gets sent back in the response, and then is checked against its original value. Is that accurate or am I maybe missing something?

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
```diff
@@ -129,9 +128,13 @@ describe('SagemakerNotebookEnvironmentLifecycleService', () => {
...
     const envHelper = new EnvironmentLifecycleHelper();
-    environment.datasetIds = ['exampleDS'];
+    environment.DATASETS = [
+      {
+        id: 'dataset-123'
```
aws-tyler, line 135: only included the `id` here and not the whole object since the rest isn't needed for this test.

workbench-core/environments/src/services/environmentService.test.ts
```diff
@@ -1038,7 +1040,8 @@ describe('EnvironmentService', () => {
         PROJ: projItem,
         ...env,
         provisionedProductId: '',
-        error: undefined
+        error: undefined,
+        datasetIds: undefined
```
maghirardelli, line 1044: Why do we need `datasetIds: undefined`?

workbench-core/environments/src/services/environmentService.ts
```diff
@@ -396,15 +394,14 @@ export class EnvironmentService {
...
-      datasetIds: params.datasetIds,
       envTypeConfigId: params.envTypeConfigId,
       updatedAt: new Date().toISOString(),
       updatedBy: user.id,
       createdAt: new Date().toISOString(),
       createdBy: user.id,
       owner: user.id,
       status: params.status || 'PENDING',
-      type: params.envTypeId,
+      type: '',
```
maghirardelli, line 404: Why don't we set this to be `${resourceTypeToKey.envType}#${params.envTypeId}${resourceTypeToKey.envTypeConfig}#${params.envTypeConfigId}` like in line 375?

workbench-core/environments/src/utilities/environmentLifecycleHelper.ts
```diff
@@ -142,13 +142,13 @@ export default class EnvironmentLifecycleHelper {
 
   /**
    * Delete the access points for all datasets attached to a given workspace.
-   * @param envMetadata - the environment for which access point(s) were created
+   * @param environment - the environment for which access point(s) were created
```
maghirardelli, line 145: environment.DATASETS does not need to be defined (see Environment type). Does this handle that case?

FernandoAranda (approved):

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.ts
```diff
@@ -32,8 +32,13 @@ export default class SagemakerNotebookEnvironmentLifecycleService implements Env
 
     const { datasetsBucketArn, mainAccountRegion, mainAccountId, mainAcctEncryptionArn } =
       await this.helper.getCfnOutputs();
+
+    const datasetIds: string[] =
+      envMetadata.DATASETS?.map((dataSet: { id: string }) => {
```
maghirardelli, line 37: Did you add a test for multiple datasets as well?

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
maghirardelli: Let's not get rid of test cases. I think this and the test above it are important. Can you fix this test to do what you think it needs to? Why does this not do what it should?

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
```diff
@@ -149,6 +152,8 @@ describe('SagemakerNotebookEnvironmentLifecycleService', () => {
     const sm = new SagemakerNotebookEnvironmentLifecycleService();
     sm.helper = envHelper;
     const response = await sm.launch(environment);
+
+    // This does not appear to be testing what the name claims it is and should be reviewed
```
pbabz, line 156: maybe preface the comment with `TODO:` ... I'm not sure how we keep track of things like this :/

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
pbabz: Not a fan of skipped tests ... Can we add a story to the backlog to come back and polish it if it's not testing the right thing?

workbench-core/environments/src/utilities/environmentLifecycleHelper.ts
maghirardelli: For consistency if you are going to rename this param, please rename the other params in this file that are used in the same way.

pbabz (approved):
Looks good!

workbench-core/environments/src/services/environmentService.test.ts
aws-tyler: Its a check to verify `datasetIds` does not appear in the final object (part of the AC)

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
aws-tyler: This is temporary and won't be getting merged in, just meant to help jump start the conversation about why this test got marked `skip` and what to do about it.

workbench-core/environments/src/utilities/environmentLifecycleHelper.ts
aws-tyler: Lodash returns `true` for `_.isEmpty(null)` and `_.isEmpty(undefined)` so we'll still get the desired behavior.

workbench-core/environments/src/services/environmentService.test.ts
maghirardelli: Would not having the line also verify that? If there is a `datasetIds` value, the `expect` would fail, right?

aws-tyler: `toEqual` only checks that there is field equality for the fields that exist. Not including a value for `datasetIds` would mean it wouldn't get checked. By defining it as `undefined`, we're explicitly saying that we want to make sure that `environment.datasetIds` is `undefined` on both objects.

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.ts
aws-tyler: I think we only need to worry about the case where its empty (which is tested in `Launch should return a mocked id` and where its non-empty. If we've got a reason to test specific sizes though just let me know and I'm happy to add it.

solutions/swb-reference/src/environment/sagemakerNotebook/sagemakerNotebookEnvironmentLifecycleService.test.ts
aws-tyler: Spoke with @SanketD92. My code changes actually make this test obsolete, which is why it didn't appear to be doing anything. Removed the test now that it isn't needed anymore.

SanketD92: Looks like we're mocking the `getDatasetsToMount` helper method in this test case and the one above to return `s3Mounts` and `iamPolicyDocument` values, when only one test case (this one) should have returned them (while the other returned empty strings).

pbabz (approved):

nguyen102 (approved):

PR #515, opened 2022-10-06
fix: environment launch metadata error handling

nguyen102:
Issue #, if available:

Description of changes:
When launching an environment, if the customer provides an invalid envTypeConfigId, projectId, or datasetIds, throw a 400 error.

Checklist:


workbench-core/environments/src/services/environmentService.test.ts
```diff
@@ -1042,5 +1042,104 @@ describe('EnvironmentService', () => {
...
+      // Get env metadata
+      const batchItems = {
+        Responses: {
+          [TABLE_NAME]: [
+            marshall({ ...projItem, resourceType: 'project' }),
+            marshall({ ...datasetItem, resourceType: 'dataset' })
+          ] // Order is important
+        }
+      };
+      // @ts-ignore
```
aws-tyler, line 1056: What is getting thrown here?

workbench-core/environments/src/services/environmentService.test.ts
```diff
@@ -1042,5 +1042,104 @@ describe('EnvironmentService', () => {
...
+        'projectId proj-123 does not exist'
+      );
+    });
+    test('failed because Dataset does not exist', async () => {
+      // BUILD
+      // Get env metadata
+      const batchItems = {
+        Responses: {
+          [TABLE_NAME]: [
+            marshall({ ...envTypeConfigItem, resourceType: 'envTypeConfig' }),
```
aws-tyler, line 1117: Looks like there might be an opportunity to do a refactor here using a `beforeEach` where you can define `batchItems`. That would prevent the copy-paste because it looks like the setup is the same and `batchItems` is the only difference. Not a blocker though.

workbench-core/environments/src/services/environmentService.test.ts
```diff
@@ -1042,5 +1042,104 @@ describe('EnvironmentService', () => {
...
+      const queryItemResponse: QueryCommandOutput = {
+        Items: envWithMetadata.map((item) => {
+          return marshall(item);
+        }),
+        $metadata: {}
+      };
+      ddbMock.on(QueryCommand).resolves(queryItemResponse);
+
+      // OPERATE && CHECK
+      await expect(envService.createEnvironment(createEnvReq, authenticateUser)).rejects.toThrow(
```
aws-tyler, line 1140: another non-blocker but something to note:  I think `toThrow` doesn't check the error code, just the error message. Might be worth investigating if Jest has a Boom extension, otherwise we can write it ourselves. thoughts?

aws-tyler (dismissed):
A couple comments but nothing blocking.

workbench-core/environments/src/services/environmentService.test.ts
nguyen102: updated

workbench-core/environments/src/services/environmentService.test.ts
nguyen102: Looks like this was not needed. Removing

workbench-core/environments/src/services/environmentService.test.ts
nguyen102: updated to expect specific boom error

aws-tyler: If you change this to `Boom.somethingElse` but leave the message the same does the test fail? I believe thats the behavior i ran into in the past; the message was the only thing being checked, not the error type.

aws-tyler (approved):

aws-tyler (approved):

PR #524, opened 2022-10-07
feat: implement listStorageLocations API

bhudnell:
Issue #, if available:

Description of changes:
Added the `listStorageLcoations` API to Datasets service.

Checklist:


workbench-core/datasets/src/schemas/createDataSet.ts
```diff
@@ -13,10 +13,11 @@ const CreateDataSetSchema: Schema = {
     datasetName: { type: 'string' },
     storageName: { type: 'string' },
     path: { type: 'string' },
-    awsAccountId: { type: 'string' }
+    awsAccountId: { type: 'string' },
+    region: { type: 'string' }
```
sevu1, line 17: Will that validate that empty string is not allowed (and should it be allowed)?

workbench-core/datasets/src/storageLocation.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface StorageLocation {
+  name: string;
+  awsAccountId?: string;
+  type?: string;
+  region?: string;
```
SanketD92, line 10: Why are `awsAccountId`, `type` and `region` optional?

workbench-core/datasets/src/ddbDataSetMetadataPlugin.ts
```diff
@@ -124,6 +127,21 @@ export class DdbDataSetMetadataPlugin implements DataSetMetadataPlugin {
     return endPointParam;
   }
 
+  public async listStorageLocations(): Promise<StorageLocation[]> {
+    const datasets = await this.listDataSets();
```
htztomic, line 131: this seems rather inefficient to query for all datasets. Keep in mind that if the return data limit exceeds 1MB, this will be missing other datasets. Revise listDataSets as it seems like it doesn't account for that. The other limitation to note here is that you can have m storage location and n amount of datasets and the time complexity grows with datasets -> O(n), not with amount of storage locations. So you will have situations where you have one storage location and 1000 datasets and you will have to look through 1000 items even though there is one storage location.

workbench-core/datasets/src/ddbDataSetMetadataPlugin.ts
```diff
@@ -199,6 +217,7 @@ export class DdbDataSetMetadataPlugin implements DataSetMetadataPlugin {
         storageName: dataSet.storageName,
         path: dataSet.path,
         awsAccountId: dataSet.awsAccountId!,
+        region: dataSet.region!,
```
htztomic, line 220: is there a reason why dataset interface keeps region as optional but you are guaranteeing it here?

htztomic (changes requested):

workbench-core/datasets/src/storageLocation.ts
bhudnell: `awsAccountId` and `region` are AWS specific parameters, and since DatasetsService isnt bound to AWS keeping them optional makes sense. `type` should be required, I will update that.

workbench-core/datasets/src/ddbDataSetMetadataPlugin.ts
bhudnell: I was following the paradigm of the other optional parameters, mainly `awsAccountId`. I will go through and check that leaving them optional doesnt break anything.

workbench-core/datasets/src/ddbDataSetMetadataPlugin.ts
bhudnell: Adding another DDB item to represent a `StorageLocation` is a good idea, but will require a larger refactor than the scope of this PR. I will make a ticket for the refactor and handle it in another PR.

workbench-core/datasets/src/schemas/createDataSet.ts
bhudnell: Empty string is and should be allowed. This schema is to validate the request body structure so no additional parameters and parameters of an incorrect type cannot be passed in.

workbench-core/datasets/src/dataSet.ts
```diff
@@ -43,4 +43,6 @@ export interface DataSet {
    * Storage Type of the DataSet
    */
   storageType?: string;
+
+  region?: string;
```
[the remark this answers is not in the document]
bhudnell: done

SanketD92 (approved):
LGTM

htztomic (approved):
LGTM

PR #530, opened 2022-10-12
feat: register internal dataset

sevu1:
Issue #, if available: GALI-1821

Description of changes:

Update dataset list page: 

image

Create new dataset:

image

image


workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+        header={
+          <Header
+            variant="h2"
+            description="Selected projects are visible to all users that are a part of the project."
+          >
+            Set permission to share project
+          </Header>
+        }
+      >
+        <TextContent>TODO: shared projects component</TextContent>
```
aws-tyler, line 235: Is this still needed?

workbench-core/datasets-ui/src/api/datasets.ts
```diff
@@ -3,16 +3,20 @@
...
+const useDatasets = (): { datasets: DatasetItem[]; areDatasetsLoading: boolean } => {
   const { data, isValidating } = useSWR(() => 'datasets', httpApiGet);
   // TODO: Once datasetService methods return response wrapped in a {data: <response>} body, replace the line below with:
   // const datasets: DatasetItem[] = data?.data ?? [];
   const datasets: DatasetItem[] = data || [];
   return { datasets, areDatasetsLoading: isValidating };
 };
 
-export { useDatasets };
+const createDataset = async (dataset: CreateDatasetForm): Promise<void> => {
```
bhudnell, line 18: Files should not be sent to the server, instead a presigned URL should be requested for each file and sent directly to the requested URL.

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+  areProjectsLoading
+}: NewDatasetFormProps): JSX.Element => {
+  const [isSubmitLoading, setIsSubmitLoading] = useState(false);
+  const [disableSubmit, setDisableSubmit] = useState(true);
+  const [error, setError] = useState('');
+  const [formData, setFormData] = useState<CreateDatasetForm>({});
+  const [formErrors, setFormErrors] = useState<CreateDatasetFormValidation>({});
+
+  const validateAll = (): boolean => {
+    let isValid = validateField<string | undefined, CreateDatasetForm, CreateDatasetFormValidation>(
```
bhudnell, line 49: instead of using let, can we return early if any of the individual validations fail?

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+}
+
+export const NewDatasetForm = ({
+  onFormSubmit,
+  projects,
+  areProjectsLoading
+}: NewDatasetFormProps): JSX.Element => {
+  const [isSubmitLoading, setIsSubmitLoading] = useState(false);
+  const [disableSubmit, setDisableSubmit] = useState(true);
+  const [error, setError] = useState('');
```
aws-tyler, line 44: Would it make sense to have the default be `null` to represent no errors at all?

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+                  value
+                );
+              }}
+              value={formData?.description || ''}
+              placeholder="Description"
+            />
+          </FormField>
+          <FormField
+            label="Owning Project"
+            // TODO: add href
```
aws-tyler, line 156: still needed?

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+export const NewDatasetForm = ({
+  onFormSubmit,
+  projects,
+  areProjectsLoading
+}: NewDatasetFormProps): JSX.Element => {
+  const [isSubmitLoading, setIsSubmitLoading] = useState(false);
+  const [disableSubmit, setDisableSubmit] = useState(true);
+  const [error, setError] = useState('');
+  const [formData, setFormData] = useState<CreateDatasetForm>({});
```
aws-tyler, line 45: This might be a "I'm still getting used to how types are used in TypeScript" thing, but it feels weird to pass in an empty object here. That only works because we've declared the fields on `CreateDatasetForm` to all be optional, correct? In practice, all of those fields are required for the request, so I don't think we want to let them be optional. Maybe we could have this be `useState<CreateDatasetForm|null>(null)` and then take away the optionality? What do you think?

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const [formErrors, setFormErrors] = useState<CreateDatasetFormValidation>({});
```
aws-tyler, line 46: same as above for optionality.

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: Idea here is to highlight all validation errors. It might be a bit confusing if user fixes field 1, clicks submit and then field 2 is showing error

workbench-core/datasets-ui/src/api/datasets.ts
sevu1: This part will be taken care with integration work. Current API does not have description (or project ids)

workbench-core/datasets-ui/src/utils.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { nameRegex, ValidationRule } from '@aws/workbench-core-swb-common-ui';
+
+export const datasetNameValidationRules: ValidationRule<string | undefined>[] = [
+  {
+    condition: (a: string | undefined) => !!a,
```
aws-tyler, line 10: maybe descriptive variable names instead of `a` to help clarify what the item being validated is?

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: Yes, that was changed quite recently after implementation started. We have another story to handle this component

aws-tyler (dismissed):

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: Yea, this is still WIP, we don't have behavior for this link

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: I'd keep empty object to avoid issues with NRE

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: I'd keep empty string to avoid issues with NRE, let say if we want to do `.toLower`

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
sevu1: I agree on comment in general, however I'd like to avoid having `| null`, otherwise we will be doing `?.` everywhere. One way to fix this is to pass empty "correct" object or convert `*Form` into class. And former might be a less of a hustle

sevu1: Updated

workbench-core/datasets-ui/src/pages/new.tsx
```diff
@@ -0,0 +1,55 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useProjects } from '@aws/workbench-core-accounts-ui';
+import { BaseLayout } from '@aws/workbench-core-swb-common-ui';
```
htztomic, line 7: wondering if workbench-core-swb-common-ui should be moved to workbench-core-common-ui. It would be better to keep this more generic

htztomic: concern is that datasets-ui now completely depends on swb-ui

workbench-core/datasets-ui/src/models/Dataset.ts
```diff
@@ -11,3 +11,17 @@ export interface DatasetItem {
   name: string;
   storageName: string;
 }
+
+export interface CreateDatasetForm {
+  name?: string;
+  description?: string;
+  projectId?: string;
```
htztomic, line 18: "project" is a concept only known to SWB

workbench-core/datasets-ui/src/components/CreateInternalDatasetForm.tsx
```diff
@@ -0,0 +1,287 @@
...
+            Upload files/folder - <i>optional</i>
+          </Header>
+        }
+      >
+        <SpaceBetween direction="vertical" size="l">
+          <FormField label="Upload files/folder" errorText={formErrors?.fileError}>
+            <FileUpload
+              data-testid="datasetFile"
+              value={formData?.file}
+              constraintText="All file type accepted. 5GB max file size."
```
htztomic, line 214:
```suggestion
              constraintText="All file types accepted. 5GB max file size."
```

workbench-core/datasets-ui/src/pages/new.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { NewDatasetForm } from '../components/CreateInternalDatasetForm';
+import { CreateDatasetForm } from '../models/Dataset';
+
+export const NewDatasetPage: NextPage = () => {
+  const router = useRouter();
+  const { projects, areProjectsLoading } = useProjects();
+
+  const breadcrumbs: BreadcrumbGroupProps.Item[] = [
+    {
+      text: 'Service Workbench',
```
htztomic, line 22: Is this meant to be a common component?

htztomic (changes requested):

workbench-core/datasets-ui/src/models/Dataset.ts
sevu1: Suggestion on better name would be appreciated

workbench-core/datasets-ui/src/pages/new.tsx
sevu1: Yes, as well as notifications. But that's out of the scope of the current PR

workbench-core/datasets-ui/src/pages/new.tsx
sevu1: These all packages are all under `workbench-core` structure. Might be okay to move to workbench-core-common-ui (in fact renaming), however out of the scope of the current PR. 
@htztomic , please elaborate on dependency between datasets-ui and swb-ui

htztomic: the BaseLayout are part of the `@aws/workbench-core-swb-common-ui`. If datasets-ui is a common component, we would prefer to keep the dependency decoupled because other MA teams will onboard, they will depend on this package which means they would also transitively depend on the swb-ui component.

workbench-core/datasets-ui/src/models/Dataset.ts
htztomic: is this suggesting that a dataset is a child to a project?

workbench-core/datasets-ui/src/pages/new.tsx
sevu1: That's true, but if you look at current list of dependencies for datasets-ui, it's already include reference to swb-common-ui. 
https://github.com/awslabs/solution-spark-on-aws/blob/develop/workbench-core/datasets-ui/package.json#L34

Confusion here could be that the name of the package should not be swb-common-ui, but common-ui, but this will be out of scope for this PR. I'll create ticket to rename swb-common-ui.

workbench-core/datasets-ui/src/models/Dataset.ts
sevu1: Currently the name is vague. But from UI, yes, this is owning project. While I can definitely rename `projectId` -> `owningProjectId`, I think you might mean more generic name rather than `project`, however I'm not sure what that name would be

htztomic: right, not all datasets will have a parent resource. I like metadata and within metadata can exist a projectId.

sevu1: Updated

htztomic (approved):

SanketD92 (approved):

PR #532, opened 2022-10-13
feat: [GALI-1786] update create hosting account

aws-tyler:
Issue #, if available: GALI-1786

Description of changes:
* Updated the request for create / update calls for hosting accounts.
  * Added `name` and `externalId`
  * Removed `environmentInstanceFiles`, `encryptionKeyArn`, `vpcId`, `subnetId`, and `cidr`
* Added logic to parse `environmentInstanceFiles` from the `artifactBucketArn`
* Added `name` to the ACC witih ACC dynamo entry
* Separated the `Create` and `Update` flows in `HostingAccountService`
  * Updated related tests. 
* Renamed `updateAccount` to `updateHostingAccountData` in `HostingAccountLifecycleService`
  * Needed to help differentiate between the `PATCH /aws-account/:id` method and the chron job that syncs hosting account data.  

Checklist:


nguyen102:
From `develop` create a feature branch name `feat/projects-api`. Then change the base of this PR to the new feature branch. Once all of the child tasks for GALI-1782 are merged into `feat/projects-api`, we'll merge `feat/projects-api` into `develop`.

nguyen102:
Did you get a chance to test if creating/updating an account through the API still propogated the `environmentInstanceFiles` and `encryptionKeyArn`? `encryptionKeyArn` should be added after the `AccountHandler` runs.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -479,6 +493,17 @@ export default class HostingAccountLifecycleService {
     return currOutputVal;
   }
 
+  private _getEnvironmentFilesPathForArn(artifactBucketArn: string): string {
+    const parsedBucketArn = artifactBucketArn.replace('arn:aws:s3::::', '').split('/');
+    const bucketName = parsedBucketArn[0];
+
+    if (_.isEmpty(bucketName) || bucketName.length === 0) {
+      throw new InvalidArtifactBucketArnError(artifactBucketArn);
```
nguyen102, line 501: Are we trying to propagate this error message to the requester? If yes, we need to use Boom. You can throw one of the boom errors from [here](https://hapi.dev/module/boom/api/?v=10.0.0#http-4xx-errors).

An example of how Boom is used is shown [here](https://github.com/awslabs/solution-spark-on-aws/blob/8b1fba979da9d3e335cb17d819cba6e2279c654a/workbench-core/accounts/src/services/projectService.ts#L61)

aws-tyler: > Did you get a chance to test if creating/updating an account through the API still propogated the `environmentInstanceFiles` and `encryptionKeyArn`? `encryptionKeyArn` should be added after the `AccountHandler` runs.

Yup, it all gets propagated to the account.

aws-tyler:
> Did you get a chance to test if creating/updating an account through the API still propogated the `environmentInstanceFiles` and `encryptionKeyArn`? `encryptionKeyArn` should be added after the `AccountHandler` runs.

Yup, it all gets propagated to the account.

aws-tyler:
Just found a bug: Updating through the API returns a 403 `User is not authorized` error. Investigating.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -14,6 +14,14 @@ import _ from 'lodash';
 import { HostingAccountStatus } from '../constants/hostingAccountStatus';
 import AccountService from '../services/accountService';
 
+export class InvalidArtifactBucketArnError extends Error {
```
nguyen102, line 17: This would be the first time we add custom error classes. Before we start adding these classes, maybe you can draft a one pager to inform the team on when we should create custom error classes and how they're used. 

Some questions I had in mind.
1. What type of errors deserve their own custom error classes?
2. What does a custom error class provide over the Boom error classes?
3. When should the custom error classes go into a specific package versus in `workbench-core-base`? Other packages might want to use the same custom error classes.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
nguyen102: We can just throw a Boom error here, and it'll bubble up to the requester. 
````
console.error(`Cannot calculate artifact bucket name from S3 ARN ${artifactBucketArn}`);
throw Boom.badImplementation('Invalid artifact bucket ARN. Cannot calculate artifact bucket name from S3 ARN); // This message will be shown to the requester. However, we don't want to provide the `artifactBucketArn` because we don't want to leak implementation details to the requester
```

workbench-core/accounts/src/services/hostingAccountService.ts
```diff
@@ -15,6 +16,28 @@ export default class HostingAccountService {
   public async create(accountMetadata: { [key: string]: string }): Promise<{ [key: string]: string }> {
     const lifecycleService = new HostingAccountLifecycleService();
 
-    return lifecycleService.initializeAccount(accountMetadata);
+    try {
```
nguyen102, line 19: We don't need a try/catch here if `initializeAccount` throws a Boom error

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
aws-tyler: Sounds good, I'll remove this for now and write something up!

bhudnell:
Looks like there is a CodeQL alert. If it is not a concern, let me know and I'll approve

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -16,8 +24,24 @@
     '/aws-accounts',
     wrapAsync(async (req: Request, res: Response) => {
       processValidatorResult(validate(req.body, CreateAccountSchema));
-      const response = await account.create(req.body);
-      res.send(response);
+      const createAccountMetadata: CreateAccountMetadata = {
+        ...req.body
+      };
+      const createdAccount = await account.create(createAccountMetadata);
+      res.send(createdAccount);
```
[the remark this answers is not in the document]
aws-tyler: @bhudnell I don't think these are actually Reflected XSS vulns. If it is, then I'm not sure how we want to handle the situation because the point of this flow is to store the data that users send us and we currently don't have any mechanisms for sanitizing that data available. Would love your thoughts on this. CC: @nguyen102 @FernandoAranda @maghirardelli

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -16,8 +25,38 @@ export function setUpAccountRoutes(router: Router, account: HostingAccountServic
...
+    })
+  );
+
+  router.patch(
+    '/aws-accounts/:id',
+    wrapAsync(async (req: Request, res: Response) => {
+      processValidatorResult(validate(req.body, UpdateAccountSchema));
+      const { name, awsAccountId, envMgmtRoleArn, hostingAccountHandlerRoleArn, externalId } = req.body;
```
aws-tyler, line 46: @bhudnell I've added what feels like extremely gratuitous escaping logic when constructing the actual `UpdateAccountMetadata` and `CreateAccountMetadata` objects from the request body in their respective methods. For some reason we're still getting pinged for the CWE. I'm pretty sure that escaping all present data mitigates the issue, right? Or is this truly a valid CWE?

Additionally, do you know any ways to better manage sanitization of an arbitrary response object? Using `escape` was the quickest thing I could think of, but it feels like it won't scale.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -479,6 +505,17 @@ export default class HostingAccountLifecycleService {
     return currOutputVal;
   }
 
+  private _getEnvironmentFilesPathForArn(artifactBucketArn: string): string {
+    const parsedBucketArn = artifactBucketArn.replace('arn:aws:s3::::', '').split('/');
+    const bucketName = parsedBucketArn[0];
+
+    if (_.isEmpty(bucketName) || bucketName.length === 0) {
+      Boom.badRequest(`Could not identify bucket name in S3 ARN ${artifactBucketArn}`);
```
nguyen102, line 513: I think you have to throw the boom error for it to propogate. Also I think this should be a `Boom.internalServerError`, since the `artifactBucketArn` is not coming from the user request, but from the CFN output. Some how the main account was set up incorrectly and outputed the wrong bucket arn for the bucket.

aws-tyler: Will do that and add a test for coverage.

solutions/swb-app/src/accountRoutes.ts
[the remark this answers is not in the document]
nguyen102: This looks like a false positive. `createdAccount` is not a user provided value. It's the value we return from our database.

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -16,8 +24,24 @@
...
+  router.patch(
+    '/aws-accounts/:id',
+    wrapAsync(async (req: Request, res: Response) => {
+      processValidatorResult(validate(req.body, UpdateAccountSchema));
+      const updateAccountMetadata: UpdateAccountMetadata = {
+        id: req.params.id,
+        ...req.body
+      };
+      const updatedAccount = await account.update(updateAccountMetadata);
+      res.send(updatedAccount);
```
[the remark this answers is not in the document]
nguyen102: This looks like a false positive. `updatedAccount` is not a user provided value. It's the value we return from our database.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
aws-tyler: Added!

solutions/swb-app/src/accountRoutes.ts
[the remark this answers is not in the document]
bhudnell: If thats the case, I think we're fine without the `escape` wrappers, unless theyre there for some other reason

solutions/swb-app/src/accountRoutes.ts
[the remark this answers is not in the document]
bhudnell: same comment as above

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -479,6 +505,17 @@ export default class HostingAccountLifecycleService {
     return currOutputVal;
   }
 
+  private _getEnvironmentFilesPathForArn(artifactBucketArn: string): string {
+    const parsedBucketArn = artifactBucketArn.replace('arn:aws:s3::::', '').split('/');
+    const bucketName = parsedBucketArn[0];
+
+    if (_.isEmpty(bucketName) || bucketName.length === 0) {
+      throw Boom.internal(`Could not identify bucket name in S3 artifact bucket ARN ${artifactBucketArn}`);
```
nguyen102, line 513: Nitpick. Let's not return the bucket arn since this is leaking implementation details to the customer. What we can do here is log the bucket name so the IT Admin can see the bucket name in the log, but not the customer. The customer won't be able to do anything with the bucket arn info anyway.

```
console.error(`Could not identify bucket name in S3 artifact bucket ARN ${artifactBucketArn}`)
throw Boom.internal('Could not identify bucket name in S3 artifact bucket');
```

aws-tyler: Updated.

nguyen102 (approved):

PR #536, opened 2022-10-13
chore: do not require account id and bucket name from user

sevu1:
Issue #, if available:

Description of changes: set account id and storage name on backend instead of user sending those parameters.

Checklist:

### DIFF
--- a/solutions/swb-app/src/datasetRoutes.ts
+++ b/solutions/swb-app/src/datasetRoutes.ts
@@ -19,7 +19,9 @@ import { processValidatorResult } from './validatorHelper';
 export function setUpDSRoutes(
   router: Router,
   dataSetService: DataSetService,
-  dataSetStoragePlugin: DataSetsStoragePlugin
+  dataSetStoragePlugin: DataSetsStoragePlugin,
+  datasetStorageAccount: string,
+  mainAccountId: string
 ): void {
   // creates new prefix in S3 (assumes S3 bucket exist already)
   router.post(
@@ -28,11 +30,12 @@ export function setUpDSRoutes(
       processValidatorResult(validate(req.body, CreateDataSetSchema));
       const dataSet = await dataSetService.provisionDataSet(
         req.body.datasetName,
-        req.body.storageName,
+        datasetStorageAccount,
         req.body.path,
-        req.body.awsAccountId,
+        mainAccountId,
         dataSetStoragePlugin
       );
+
       res.status(201).send(dataSet);
     })
   );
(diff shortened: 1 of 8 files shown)


bhudnell:
run a git pull from develop for the `feat/swb-byob` branch. There have been changes to the datasets package that may affect this PR. Mainly adding the region field to the create/importDataset functions

bhudnell (commented):
It also is worth discussing moving the schema objects into `swb-app` since they are defining API request schema

SanketD92 (dismissed):

solutions/swb-app/src/datasetRoutes.ts
```diff
@@ -44,9 +47,9 @@ export function setUpDSRoutes(
       processValidatorResult(validate(req.body, CreateDataSetSchema));
```
bhudnell, line 47: should one of these routes still take `storageName` and `awsAccountId` for external datasets? or will those be different routes?

sevu1: I'd prefer to keep routes separate is it would be a bit tricky to manage + run validation; additionally they are different entities, we might rename this API to reflect that, e.g. `datasets/internal`. @SanketD92 , thoughts ?

workbench-core/datasets/src/schemas/createDataSet.ts
```diff
@@ -11,13 +11,11 @@ const CreateDataSetSchema: Schema = {
   type: 'object',
```
bhudnell, line 11: Has there been discussion about moving the schema objects into `swb-app` since they are SWB specific?

sevu1: Not yet. @bhudnell , you will be the part of those discussions for sure

solutions/swb-app/src/datasetRoutes.ts
SanketD92: Initially we were planning to use the `datasets/import` API for external datasets, but it makes sense to separate our APIs for internal and external (the existing ones can stay as-is for internal, no need for rename). For external DS, we should be able to specify the dataset's (non-main) account ID and bucket name (after it has been registered) for importing it as an external DS

SanketD92: @bhudnell the answer to your question is based on whether `datasetService.importDataSet()` acts as the entry point for registering both internal and external datasets.

workbench-core/datasets/src/schemas/createDataSet.ts
sevu1: Resolving as it's being discussed offline

SanketD92 (approved):

PR #549, opened 2022-10-18
feat: mock dataset service response and e2e create internal dataset flow

sevu1:
Issue #, if available: GALI-1876

Description of changes:

  e2e create internal dataset flow. Pass ui payload to backend; mock temporary unavailable dataset service response - echo back ui params; negative dataset tests

Checklist:

### DIFF
--- a/solutions/swb-reference/integration-tests/support/clientSession.ts
+++ b/solutions/swb-reference/integration-tests/support/clientSession.ts
@@ -52,8 +52,11 @@ export default class ClientSession {
       },
       function (error: AxiosError) {
         if (error.response) {
+          console.error(error.response.data);
           return Promise.reject(new HttpError(error.response.status, error.response.data));
         }
+
+        console.error(error);
         return Promise.reject(error);
       }
     );
--- a/solutions/swb-reference/integration-tests/support/resources/datasets/datasets.ts
+++ b/solutions/swb-reference/integration-tests/support/resources/datasets/datasets.ts
@@ -28,17 +28,17 @@ export default class Datasets extends CollectionResource {
     return {
       datasetName: resource.datasetName ?? dataSetName,
       path: resource.path ?? dataSetName,
-      storageName: resource.storageName,
-      awsAccountId: resource.awsAccountId,
-      region: resource.region
+      description: resource.description,
+      customMetadata: resource.customMetadata
     };
   }
 }
 
 interface DataSetCreateRequest {
   datasetName: string;
-  storageName: string;
   path: string;
-  awsAccountId: string;
-  region: string;
+  description?: string;
+  customMetadata: {
+    owningProjectId: string;
+  };
 }
--- a/solutions/swb-reference/integration-tests/support/utils/utilities.ts
+++ b/solutions/swb-reference/integration-tests/support/utils/utilities.ts
@@ -8,6 +8,14 @@ import { DEFAULT_POLLING_INTERVAL_SECONDS, DEFAULT_POLLING_MAX_WAITING_SECONDS }
 import HttpError from './HttpError';
 
 /**
+ * Marks all properties (including nested) as Partial.
+ * https://stackoverflow.com/questions/47914536/use-partial-in-nested-property-with-typescript
+ */
+type RecursivePartial<T> = {
+  [P in keyof T]?: RecursivePartial<T[P]>;
+};
+
+/**
  * Returns a promise that will be resolved in the requested time, ms.
  * Example: await sleep(200);
  * https://stackoverflow.com/questions/951021/what-is-the-javascript-version-of-sleep/39914235#39914235
@@ -59,4 +67,4 @@ async function poll<T>(
   }
 }
 
-export { sleep, checkHttpError, poll, getFakeEnvId };
+export { sleep, checkHttpError, poll, getFakeEnvId, RecursivePartial };
--- a/workbench-core/datasets/src/schemas/createDataSet.ts
+++ b/workbench-core/datasets/src/schemas/createDataSet.ts
@@ -12,10 +12,18 @@ const CreateDataSetSchema: Schema = {
   properties: {
     datasetName: { type: 'string' },
     path: { type: 'string' },
-    region: { type: 'string' }
+    description: { type: 'string' },
+    customMetadata: {
+      type: 'object',
+      properties: {
+        owningProjectId: { type: 'string' }
+      },
+      additionalProperties: false,
+      required: ['owningProjectId']
+    }
   },
   additionalProperties: false,
-  required: ['datasetName', 'path', 'region']
+  required: ['datasetName', 'path', 'customMetadata']
 };
 
 export default CreateDataSetSchema;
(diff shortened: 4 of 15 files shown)


solutions/swb-reference/integration-tests/support/clientSession.ts
```diff
@@ -52,8 +52,11 @@ export default class ClientSession {
       },
       function (error: AxiosError) {
         if (error.response) {
+          console.error(error.response.data);
           return Promise.reject(new HttpError(error.response.status, error.response.data));
         }
+
+        console.error(error);
```
bhudnell, line 59: should these log statements be here?

workbench-core/datasets/src/schemas/createDataSet.ts
```diff
@@ -12,10 +12,18 @@ const CreateDataSetSchema: Schema = {
   properties: {
     datasetName: { type: 'string' },
     path: { type: 'string' },
-    region: { type: 'string' }
+    description: { type: 'string' },
+    customMetadata: {
+      type: 'object',
+      properties: {
+        owningProjectId: { type: 'string' }
```
DaveMadsen, line 19: the reference to "owningProjectId" does not belong here. If we define it here, the metadata is prescribed, not custom. It is fine to extend this within Service Workbench and define it there as it is metadata which is custom to the service workbench product or the projects component.

solutions/swb-reference/integration-tests/support/clientSession.ts
sevu1: Yes, this is intentional. In tests I saw only promise rejection, but no details. This helped a lot in debugging the issue

workbench-core/datasets/src/models/dataSetType.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export enum DataSetType {
+  internal = 'internal',
+  external = 'external'
```
bhudnell, line 8:
```suggestion
  INTERNAL = 'internal',
  EXTERNAL = 'external'
```

workbench-core/datasets/src/schemas/createDataSet.ts
sevu1: There was already a suggestion to move schema to api level, but that was not finalized. I also feel that schema just does not belong to core. This is used for API validation and purely relates to SWB. We still can have any shape of metadata in dataset service though

workbench-core/datasets/src/schemas/createDataSet.ts
```diff
@@ -12,10 +12,18 @@ const CreateDataSetSchema: Schema = {
...
+      },
+      additionalProperties: false,
+      required: ['owningProjectId']
+    }
   },
   additionalProperties: false,
-  required: ['datasetName', 'path', 'region']
+  required: ['datasetName', 'path', 'customMetadata']
```
bhudnell, line 26: `customMetadata` should not be required. If a user doesn't have custom metadata to pass in, they shouldnt be required to pass in an empty object

sevu1: But we still have owning project here. Possibly I'd need to move project under main entity

workbench-core/datasets/src/schemas/createDataSet.ts
DaveMadsen: Can we move this file out of the `workbench-core/datasets`? This will be shipped as a standalone NPM separate for SWB. It should not contain a pure SWB construct.

nguyen102: After giving this further though, I think we can move the schema to `swb-app`. We can put all of the schema under the schema folder. Once the request is received we should validate it's in the correct shape to be passed to the workbench core services.

Also remember that in the request we can call the field `owningProjectId`, but when that field is written to DDB it should be written as `owner`. This is so we can use the preexisting DDB index `getResourceByOwner`

solutions/swb-reference/integration-tests/support/resources/datasets/datasets.ts
```diff
@@ -28,17 +28,17 @@ export default class Datasets extends CollectionResource {
...
 }
 
 interface DataSetCreateRequest {
   datasetName: string;
-  storageName: string;
   path: string;
-  awsAccountId: string;
-  region: string;
+  description?: string;
+  customMetadata: {
```
SanketD92, line 41: why `customMetadata.owningProjectId` and not just `owner` (similar to what the actual API expects)?

solutions/swb-reference/integration-tests/support/utils/utilities.ts
```diff
@@ -7,6 +7,14 @@ import { v4 as uuidv4 } from 'uuid';
 import { DEFAULT_POLLING_INTERVAL_SECONDS, DEFAULT_POLLING_MAX_WAITING_SECONDS } from './constants';
 import HttpError from './HttpError';
 
+/**
+ * Marks all properties (including nested) as Partial.
+ * https://stackoverflow.com/questions/47914536/use-partial-in-nested-property-with-typescript
+ */
+type RecursivePartial<T> = {
```
SanketD92, line 14: Could you be more descriptive in tsdocs as to why this is needed/preferred as compared to previous types?

workbench-core/datasets/src/models/dataSetType.ts
```diff
@@ -0,0 +1,9 @@
...
+}
```
SanketD92, line 9: Nice!

workbench-core/datasets/src/schemas/createDataSet.ts
```diff
@@ -12,10 +12,18 @@ const CreateDataSetSchema: Schema = {
   properties: {
     datasetName: { type: 'string' },
     path: { type: 'string' },
-    region: { type: 'string' }
+    description: { type: 'string' },
+    customMetadata: {
```
SanketD92, line 16: Are we expecting other metadata to be provided during the create call? Any downside in replacing `customMetadata` object with a string attribute `owner`?

solutions/swb-reference/integration-tests/support/resources/datasets/datasets.ts
sevu1: It makes more sense for UI to call it owningProjectId, but for more generic dataset service we would keep it as owner

solutions/swb-reference/integration-tests/support/utils/utilities.ts
sevu1: That's being removed

workbench-core/datasets/src/schemas/createDataSet.ts
sevu1: Done, but property name is owningProjectId

bhudnell (approved):

PR #550, opened 2022-10-18
feat: add optional dataset fields

bhudnell:
Issue #, if available:

Description of changes:
added optional `owner`, `type`, and `description` fields to the `DataSet` interface

Checklist:


workbench-core/datasets/src/dataSet.ts
```diff
@@ -19,6 +19,21 @@ export interface DataSet {
...
 
+  /**
+   * (optional) a description of the dataset
+   */
+  description?: string;
+
+  /**
+   * (optional) the owner of the dataset
+   */
+  owner?: string;
```
SanketD92, line 30: @sevu1 Do we want this to be `owner` or `owningProject`?

SanketD92 (dismissed):

workbench-core/datasets/src/dataSet.ts
```diff
@@ -19,6 +19,21 @@ export interface DataSet {
...
+
+  /**
+   * (optional) the type of the dataset
+   */
+  type?: string;
```
sevu1, line 35: Should we use enum here?

bhudnell: We dont want to restrict type to a defined enum. This still works though if SWB wants to use an enum internally:
```ts
enum TestEnum {
  STRING_ONE = 'one',
  STRING_TWO = 'two
}

const testString: string = TestEnum.STRING_ONE;
```

sevu1 (approved):

PR #556, opened 2022-10-19
feat(workbench-core/authentication): updated User interface to include status

bhudnell:
Issue #, if available:

Description of changes:
- updated User interface to include status, updated unit tests
- replaced User with CreateUser in createUser function
- removed viperlight scan from pre-commit hook as it does not support Windows

Checklist:


DaveMadsen (approved):

insignias (approved):

workbench-core/authentication/src/userManagementService.test.ts
```diff
@@ -3,63 +3,58 @@
...
+  const userManagementService: UserManagementService = new UserManagementService(mockUserManagementPlugin);
+
   beforeEach(() => {
     jest.clearAllMocks();
   });
 
   test('getUser', async () => {
-    await userManagementService.getUser(mockUid);
-    expect(mockUserManagementPlugin.getUser).toBeCalledTimes(1);
-    expect(mockUserManagementPlugin.getUser).toBeCalledWith(mockUid);
```
kpark277, line 44: Why get rid of these lines in all the tests? I think it's worthwhile to verify that the plugin is being called correctly.

bhudnell: Its testing the underlying implementation of the function. If we were to change how this function works under the hood this test would most likely fail even if the inputs and outputs are the same.

bhudnell: I changed the tests to treat the function as a black box and only test the inputs and outputs.

kpark277 (approved):

PR #560, opened 2022-10-20
feat: [GALI-1784] GET /aws-account/:id

aws-tyler:
Issue #, if available:
GALI-1784

Description of changes:
Added support and tests for getting individual accounts with and without metadata

Checklist:


workbench-core/accounts/src/services/accountService.ts
```diff
@@ -175,4 +177,58 @@ export default class AccountService {
...
+
+    if (!accountEntry.Item) {
+      throw Boom.notFound(`Could not find account ${accountId}`);
+    }
+
+    return accountEntry.Item! as unknown as Account;
+  }
+
+  private async _getAccountWithMetadata(accountId: string): Promise<Account> {
+    const pk = `ACC#${accountId}`;
```
FernandoAranda, line 194: use `resourceTypeKey.account`

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -175,4 +177,58 @@ export default class AccountService {
 
     return accountMetadata.id;
   }
+
+  private async _getAccountWithoutMetadata(accountId: string): Promise<Account> {
+    const accountEntry = (await this._aws.helpers.ddb
+      .get({ pk: `${resourceTypeToKey.account}#${accountId}` })
+      .execute()) as GetItemCommandOutput;
+
+    if (!accountEntry.Item) {
```
FernandoAranda, line 186: does the `ddb.get` always returns a value? Do we need a validation for `accountEntry` like `accountEntry?.item`

pbabz (dismissed):

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -175,4 +177,58 @@ export default class AccountService {
...
+      // parent environment item
+      const sk = item.sk as unknown as string;
+
+      if (sk === pk) {
+        account = { ...account, ...item };
+        continue;
+      }
+
+      const associationPrefix = sk.split('#')[0];
+
```
FernandoAranda, line 227: can we add this prefix to `resourcetypeKey.CostCenter` if it doesnt exist and use it instead

workbench-core/accounts/src/services/accountService.ts
aws-tyler: I believe accountEntry is always valid, but `Item` can be `any | undefined`. I'll double check though.

aws-tyler: The type checker confirmed that `accountEntry` should always be present.

workbench-core/accounts/src/services/accountService.ts
aws-tyler: fixed

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -20,7 +20,13 @@ import { wrapAsync } from './errorHandlers';
 import { processValidatorResult } from './validatorHelper';
 
 export function setUpAccountRoutes(router: Router, account: HostingAccountService): void {
-  // Provision
+  router.get(
+    '/aws-accounts/:id',
+    wrapAsync(async (req: Request, res: Response) => {
+      res.send(await account.get(req.params.id));
```
sevu1, line 26: What do we get if account not found? undefined?

aws-tyler: Account IDs that don't match any account currently return a `403 User not authorized`. We could return a 404, but 403s are nice because it prevents anyone from spamming UUIDs to find out which ones are valid and which ones aren't: you will only see the data you're authorized to see.

aws-tyler: Idk if thats _why_ we're returning a 403, but its a nice security trick 🤷 .

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -9,20 +9,28 @@ import Boom from '@hapi/boom';
 import _ from 'lodash';
 import { HostingAccountStatus } from '../constants/hostingAccountStatus';
 
-interface Account {
-  id: string | undefined;
+export interface CostCenter {
```
sevu1, line 12: Wonder if we should consider move this and interface below under ``models` folder

aws-tyler: For now I left that to be decided when we start the `CostCenter` epic, but ya i agree, i just don't know where that folder will end up being and i figured it'd be easier to just do it when we start building out cost center infrastructure.

solutions/swb-app/src/accountRoutes.ts
sevu1: I guess something might be broken. I'd expect 404: https://github.com/aws-solutions/solution-spark-on-aws/pull/560/files#diff-896fb3fe131fd2efa124d3ab55693f1ddb4133831a74e07b641617ec0d812bccR199 . Could you please double check?

FernandoAranda (approved):

solutions/swb-app/src/accountRoutes.ts
aws-tyler: @sevu1 We've discovered an interesting bug! When you put in an invalid UUID as the account id, a 403 is returned, but if you put in a valid UUID that does not match any accounts, it returns a 404! @nguyen102 I'm guessing this isn't the expected behavior?

aws-tyler: I'm going to merge the change since the bug belongs to the routing code and not my changes .I'm guessing we'll want to file a ticket to change it so that any invalid id returns a 404? (or a 403 if we don't want to let attackers try and guess UUIDs?)

sevu1: It could be 400 as well

PR #576, opened 2022-10-27
feat(environment type filter): add dynamic operators capability to en…

FernandoAranda:
Issue #, if available:
GALI-1841

Description of changes:
Add dynamic operators capability and sorting to Environment types List API

Checklist:

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -799,6 +799,7 @@ importers:
       '@aws/eslint-config-workbench-core-eslint-custom': workspace:*
       '@aws/workbench-core-audit': workspace:*
       '@aws/workbench-core-logging': workspace:*
+      '@hapi/boom': ^10.0.0
       '@rushstack/eslint-config': ^3.0.0
       '@rushstack/heft': ^0.47.5
       '@rushstack/heft-jest-plugin': ^0.3.28
@@ -816,6 +817,7 @@ importers:
       istanbul-badges-readme: 1.8.1
       jest: ^27.5.1
       js-yaml: ^4.1.0
+      jsonschema: ^1.4.1
       license-check-and-add: ^4.0.5
       license-checker: ^25.0.1
       lodash: ^4.17.21
@@ -824,8 +826,8 @@ importers:
       sort-package-json: ^1.57.0
       ts-jest: ^27.1.3
       typescript: ^4.5.2
-      yaml-cfn: ^0.3.1
       uuid: ^8.3.2
+      yaml-cfn: ^0.3.1
     dependencies:
       '@aws-cdk/aws-dynamodb': 1.158.0
       '@aws-sdk/client-cloudformation': 3.130.0
@@ -846,8 +848,10 @@ importers:
       '@aws-sdk/util-dynamodb': 3.105.0
       '@aws/workbench-core-audit': link:../audit
       '@aws/workbench-core-logging': link:../logging
+      '@hapi/boom': 10.0.0
       aws-sdk-client-mock: 0.6.2_e3d47e7de0bdd7e8ce04c01f4d10cebe
       js-yaml: 4.1.0
+      jsonschema: 1.4.1
       lodash: 4.17.21
       uuid: 8.3.2
     devDependencies:
--- a/workbench-core/base/src/interfaces/FilterRequest.ts
+++ b/workbench-core/base/src/interfaces/FilterRequest.ts
@@ -7,5 +7,5 @@ import { QueryParameterFilter } from './queryParameterFilter';
 
 export interface FilterRequest {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  [key: string]: QueryParameterFilter<any> | undefined;
+  [key: string]: QueryParameterFilter<unknown> | undefined;
 }
--- a/workbench-core/base/src/interfaces/FilterRequest.ts
+++ b/workbench-core/base/src/interfaces/FilterRequest.ts
@@ -1,0 +1,11 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { QueryParameterFilter } from './queryParameterFilter';
+
+export interface FilterRequest {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: QueryParameterFilter<any> | undefined;
+}
--- a/workbench-core/base/src/interfaces/queryParameterFilter.ts
+++ b/workbench-core/base/src/interfaces/queryParameterFilter.ts
@@ -5,7 +5,7 @@
 
 /************
  *
- * Only one operator can be defined by property, if multiple operators are dfined, dynamo service will throw an exception.
+ * Only one operator can be defined by property, if multiple operators are defined, dynamo service will throw an exception.
  *
  ************/
 export interface QueryParameterFilter<T> {
--- a/workbench-core/base/src/interfaces/queryParameterFilter.ts
+++ b/workbench-core/base/src/interfaces/queryParameterFilter.ts
@@ -3,6 +3,11 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
+/************
+ *
+ * Only one operator can be defined by property, if multiple operators are dfined, dynamo service will throw an exception.
+ *
+ ************/
 export interface QueryParameterFilter<T> {
   eq?: T;
   lt?: T;
--- a/workbench-core/base/src/interfaces/queryParameterFilter.ts
+++ b/workbench-core/base/src/interfaces/queryParameterFilter.ts
@@ -1,0 +1,14 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface QueryParameterFilter<T> {
+  eq?: T;
+  lt?: T;
+  lte?: T;
+  gt?: T;
+  gte?: T;
+  between?: { value1: T; value2: T };
+  begins?: T;
+}
--- a/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
+++ b/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
@@ -12,7 +12,7 @@ describe('Query params filters utils', () => {
   describe('validateSingleSortAndFilter', () => {
     test('should throw exception when request has filter and sorting', () => {
       expect(() => validateSingleSortAndFilter({ name: { eq: 'Sauron' } }, { status: 'desc' })).toThrow(
-        'Cannot apply a filter and sort at the same time'
+        'Cannot apply a filter and sort to different properties at the same time'
       );
     });
 
--- a/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
+++ b/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
@@ -28,7 +28,7 @@ describe('Query params filters utils', () => {
       );
     });
 
-    test('sohuld not throw exception when there is only one filter', () => {
+    test('should not throw exception when there is only one filter', () => {
       expect(() => validateSingleSortAndFilter({ name: { eq: 'Sauron' } }, undefined)).not.toThrow();
     });
 
--- a/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
+++ b/workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
@@ -3,8 +3,8 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 import {
-  parseQueryParamFilter,
-  parseQueryParamSort,
+  getFilterQueryParams,
+  getSortQueryParams,
   validateSingleSortAndFilter
 } from './queryParamsFiltersUtils';
 
@@ -37,84 +37,70 @@ describe('Query params filters utils', () => {
     });
   });
 
-  describe('parseQueryParamFilter', () => {
+  describe('getFilterQueryParams', () => {
     test('should return undefined when filter is not defined', () => {
-      expect(parseQueryParamFilter(undefined, 'sortTest', 'sortGSI')).toEqual(undefined);
+      expect(getFilterQueryParams(undefined, [])).toEqual({});
     });
 
     test('should parse successfuly a between queryParam', () => {
       const inputQueryParam = {
-        between: {
-          value1: 1,
-          value2: 10
+        age: {
+          between: {
+            value1: 1,
+            value2: 10
+          }
         }
       };
       const expectedResult = {
-        index: 'testGSI',
-        sortKey: 'testProp',
-        between: { value1: { N: 1 }, value2: { N: 10 } }
+        index: 'getResourceByAge',
+        sortKey: 'age',
+        between: {
+          value1: { N: '1' },
+          value2: { N: '10' }
+        }
       };
-      expect(parseQueryParamFilter(inputQueryParam, expectedResult.sortKey, expectedResult.index)).toEqual(
-        expectedResult
-      );
+      expect(getFilterQueryParams(inputQueryParam, ['getResourceByAge'])).toEqual(expectedResult);
     });
 
-    test('should parse successfuly a eq queryParam', () => {
-      const inputQueryParam = {
-        eq: 'Sauron'
-      };
-      const expectedResult = {
-        index: 'testGSI',
-        sortKey: 'testProp',
-        eq: { S: 'Sauron' }
-      };
-      expect(parseQueryParamFilter(inputQueryParam, expectedResult.sortKey, expectedResult.index)).toEqual(
-        expectedResult
-      );
-    });
     test.each(['eq', 'lt', 'lte', 'gt', 'gte', 'begins'])(
       'should parse successfully a %p query param',
       (a) => {
         const inputQueryParam = {
-          [a]: 'Sauron'
+          name: {
+            [a]: 'Sauron'
+          }
         };
         const expectedResult = {
-          index: 'testGSI',
-          sortKey: 'testProp',
+          index: 'getResourceByName',
+          sortKey: 'name',
           [a]: { S: 'Sauron' }
         };
-        expect(parseQueryParamFilter(inputQueryParam, expectedResult.sortKey, expectedResult.index)).toEqual(
-          expectedResult
-        );
+        expect(getFilterQueryParams(inputQueryParam, ['getResourceByName'])).toEqual(expectedResult);
       }
     );
   });
 
-  describe('parseQueryParamSort', () => {
+  describe('getSortQueryParams', () => {
     test('should return undefined when sort is not defined', () => {
-      expect(parseQueryParamSort(undefined, 'sortTest', 'sortGSI')).toEqual(undefined);
+      expect(getSortQueryParams(undefined, [])).toEqual({});
     });
 
     test('should parse successfuly an ascending sort', () => {
       const expectedResult = {
-        index: 'testGSI',
-        sortKey: 'testProp',
+        index: 'getResourceByName',
+        sortKey: 'name',
         forward: true
       };
-      expect(parseQueryParamSort('asc', expectedResult.sortKey, expectedResult.index)).toEqual(
-        expectedResult
-      );
+      expect(getSortQueryParams({ name: 'asc' }, ['getResourceByName'])).toEqual(expectedResult);
     });
 
     test('should parse successfuly a descending sort', () => {
       const expectedResult = {
-        index: 'testGSI',
-        sortKey: 'testProp',
+        index: 'getResourceByName',
+        sortKey: 'name',
         forward: false
       };
-      expect(parseQueryParamSort('desc', expectedResult.sortKey, expectedResult.index)).toEqual(
-        expectedResult
-      );
+      expect(getSortQueryParams({ name: 'desc' }, ['getResourceByName'])).toEqual(expectedResult);
     });
   });
 });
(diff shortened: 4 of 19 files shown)


common/config/rush/pnpm-lock.yaml
```diff
@@ -799,6 +799,7 @@ importers:
       '@aws/eslint-config-workbench-core-eslint-custom': workspace:*
       '@aws/workbench-core-audit': workspace:*
       '@aws/workbench-core-logging': workspace:*
+      '@hapi/boom': ^10.0.0
```
aws-tyler, line 802: Don't we already have `boom` in the project? Is this a major version upgrade?

workbench-core/environments/src/services/environmentTypeService.test.ts
```diff
@@ -97,23 +97,173 @@ describe('environmentTypeService', () => {
...
+        filter: { status: { gte: 'A' } }
+      });
+
+      // CHECK
+      expect(actualResponse).toEqual({ data: [envType] });
+    });
+
+    test('sort by asc status successfully', async () => {
+      // BUILD
+      const queryItemResponse: QueryCommandOutput = {
```
aws-tyler, line 205: Not blocking: Jest provides a really neat way of letting you run setups within a given context so you don't have to copy and paste as much. It makes the tests cleaner and easier to read. [Here's an example of how its done](https://github.com/aws-solutions/solution-spark-on-aws/blob/develop/workbench-core/accounts/src/services/costCenterService.test.ts#L118). Let me know if you want to chat.

workbench-core/base/src/interfaces/queryParameterFilter.ts
```diff
@@ -0,0 +1,14 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface QueryParameterFilter<T> {
+  eq?: T;
+  lt?: T;
+  lte?: T;
+  gt?: T;
+  gte?: T;
+  between?: { value1: T; value2: T };
```
aws-tyler, line 12: maybe `start` and `end` to indicate a range?

workbench-core/base/src/utilities/queryParamsFiltersUtils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      ).toThrow('Cannot apply more than one filter.');
+    });
+
+    test('should throw exception when there is more than one sort', () => {
+      expect(() => validateSingleSortAndFilter(undefined, { status: 'desc', name: 'asc' })).toThrow(
+        'Cannot sort by more than one attribute'
+      );
+    });
+
+    test('sohuld not throw exception when there is only one filter', () => {
```
aws-tyler, line 31: small typo `sohuld`

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Parses Object with format \{ field: \{ operator: value \} \} into QueryParam object consumed by dynamoDB
+ *
+ * @param filter - filter request object field
+ * @param sortKey - field name in dynamoDB
+ * @param gsi - gsi to use for filter
+ *
+ * @returns QueryParams object
+ ************************************************************/
+export function parseQueryParamFilter<T>(
+  filter: QueryParameterFilter<T> | undefined,
```
aws-tyler, line 40: I think you can put `filter?: QueryParameterFilter<T>` and it will automatically union the type with `undefined`

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  sortKey: string,
+  gsi: string
+): QueryParams | undefined {
+  if (!filter || Object.keys(filter).length === 0 || Object.keys(filter).length > 1) {
+    return undefined;
```
aws-tyler, line 45: Would it be worth throwing an error here to indicate what went wrong with the parsing?

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+  if (filter?.between && filter.between.value1 && filter.between.value2) {
```
aws-tyler, line 47: I think the `value` checks should be inside of the body of the `filter?.between` check, right? If we have a between but don't have both values, I think we'd probably want to throw an error?

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      index: gsi,
+      sortKey: sortKey,
+      between: { value1, value2 }
+    };
+  }
+  const filterOperator: keyof QueryParameterFilter<T> = Object.keys(
+    filter
+  )[0] as keyof QueryParameterFilter<T>;
+  //eslint-disable-next-line security/detect-object-injection
```
aws-tyler, line 60: is there a way of doing this without disabling the security check for [Object Injection](https://github.com/nodesecurity/eslint-plugin-security/blob/main/docs/the-dangers-of-square-bracket-notation.md)?

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+ ************************************************************/
+export function parseQueryParamSort(
+  sort: 'asc' | 'desc' | undefined,
+  sortKey: string,
+  gsi: string
+): QueryParams | undefined {
+  if (!sort) {
+    return undefined;
+  }
+  const forward: boolean = sort !== 'desc';
```
aws-tyler, line 90: Once the TS is transpiled to JS, would it be possible for someone to send in a value that wasn't `asc`, `desc`, or `undefined` via a query param? I'm worried that only checking for `desc` means that at runtime someone could still send in something invalid.

workbench-core/environments/src/services/environmentTypeService.ts
```diff
@@ -49,20 +57,43 @@ export default class EnvironmentTypeService {
...
+        filter[filterProperty],
+        filterProperty,
+        `getResourceBy${gsiFilterProperty}`
+      );
+      queryParams = { ...queryParams, ...queryParamsFilter };
+    }
+    if (sort) {
+      const sortProperty = Object.keys(sort)[0] as keyof EnvironmentTypeSort;
+      const gsiSortProperty = gsiPropertyNames.filter((prop) => prop.toLowerCase() === sortProperty)[0];
+      //eslint-disable-next-line security/detect-object-injection
```
aws-tyler, line 89: Can we do this without enabling [Object Injection](https://github.com/nodesecurity/eslint-plugin-security/blob/main/docs/the-dangers-of-square-bracket-notation.md)?

aws-tyler (changes requested):
I'm happy to approve as soon as we know if we can prevent the Object Injection or not. Only blocking because that could be a potential security issue. Otherwise everything looks great!

common/config/rush/pnpm-lock.yaml
FernandoAranda: we dont have it in base, only in other packages

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: the api is able to limit those values with the scema, on this layer I will just asume that if not asc or desc default is asc in the new commit I will add

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: fixed in next commit

workbench-core/environments/src/services/environmentTypeService.ts
FernandoAranda: fixed in next commit

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: fixed in next commit

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: fixed in commit

workbench-core/base/src/interfaces/queryParameterFilter.ts
FernandoAranda: changing the values here implies we need to do a parse of fields when doing the marshall to the object, so I felt this was more convenient

aws-tyler (approved):

workbench-core/environments/src/interfaces/listEnvironmentTypeRequest.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { QueryParameterFilter, FilterRequest, SortRequest } from '@aws/workbench-core-base';
+
+export interface ListEnvironmentTypeRequest {
```
maghirardelli, line 8: nit: should environment type(S) be plural?

workbench-core/environments/src/interfaces/listEnvironmentTypeRequest.ts
```diff
@@ -0,0 +1,23 @@
...
+  filter?: EnvironmentTypeFilter;
+  sort?: EnvironmentTypeSort;
+}
+
+export interface EnvironmentTypeFilter extends FilterRequest {
+  status?: QueryParameterFilter<string>;
+  name?: QueryParameterFilter<string>;
+}
+
+export interface EnvironmentTypeSort extends SortRequest {
```
maghirardelli, line 20: this is so cool

workbench-core/environments/src/schemas/listEnvironmentType.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+// Schema for listEnvironmentType API
+import { QueryParameterFilterSchema } from '@aws/workbench-core-base';
+import { Schema } from 'jsonschema';
+
+const ListEnvironmentTypeSchema: Schema = {
```
maghirardelli, line 10: nit: plural for list

workbench-core/environments/src/schemas/listEnvironmentType.ts
```diff
@@ -0,0 +1,36 @@
...
+  id: '/listEnvironmentType',
```
maghirardelli, line 11: nit: plural

workbench-core/environments/src/services/environmentTypeService.ts
```diff
@@ -49,20 +53,24 @@ export default class EnvironmentTypeService {
...
+    validateSingleSortAndFilter(filter, sort);
     let queryParams: QueryParams = {
       key: { name: 'resourceType', value: this._resourceType },
       index: 'getResourceByCreatedAt',
       limit: pageSize && pageSize >= 0 ? pageSize : DEFAULT_API_PAGE_SIZE
     };
+    const gsiNames = ['getResourceByName', 'getResourceByStatus'];
+    const filterQuery = getFilterQueryParams(filter, gsiNames);
+    const sortQuery = getSortQueryParams(sort, gsiNames);
+    queryParams = { ...queryParams, ...filterQuery, ...sortQuery };
```
maghirardelli, line 73: I assume `filterQuery` and `sortQuery` will never both be defined? Can we add a check for that?

maghirardelli: Also this is so neat I love it.

FernandoAranda: the validation is applied before this, so at this point it should have a valid one, I also added a check to allow same field for sort and filter, I even added at some point multiple filter to one field but realized that our dynamo layer doesn't allow that

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Check that at most one sort attribute is defined because we not support sorting by more than one attribute
+  if (sortAttributesLength > 1) {
+    throw Boom.badRequest('Cannot sort by more than one attribute.');
+  }
+  const hasSameProperty =
+    filterAttributesLength === 1 &&
+    sortAttributesLength === 1 &&
+    Object.keys(filter ?? {})[0] === Object.keys(sort ?? {})[0];
+  if (filter && sort && !hasSameProperty) {
+    throw Boom.badRequest('Cannot apply a filter and sort at the same time');
```
bhudnell, line 32: It looks like you can apply a filter and a stort at the same time as long as they're operating on the same property?

FernandoAranda: yes, dynamo allows this , I tested it and works

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    return queryParams;
+  }
+  if (Object.keys(filter).length > 1) {
+    throw Boom.badRequest('Cannot filter by more than one attribute.');
+  }
+  //only one entry will exist due to validation above
+  Object.entries(filter).forEach(([key, value]) => {
+    //transform { operator: value } into { operator { type: value } }
+    const filterQueryParam = (
+      value?.between ? { between: marshall(value.between) } : marshall(value)
```
bhudnell, line 57: this line make me think that the `QueryParameterFilter` can either have `between` or the rest of the fields (`lt`, `lte`, `eq`, etc) defined. Is that correct?

FernandoAranda: well, we can only have one type of filter, dynamo will complain if we have more than one, but I had a problem applying marshall to between, I will add an extra type , so I had to send only what is inside between to make it work , other operators can just be marshall and will work

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+  let queryParams: QueryParams = {};
+  //validate filter request doesnt have more than one filter or no filters
+  if (!filter || Object.keys(filter).length === 0) {
+    return queryParams;
+  }
+  if (Object.keys(filter).length > 1) {
+    throw Boom.badRequest('Cannot filter by more than one attribute.');
+  }
+  //only one entry will exist due to validation above
+  Object.entries(filter).forEach(([key, value]) => {
```
bhudnell, line 54: Since only one entry is allowed I feel like this is adding unnecessary complication to the function.

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+export function getSortQueryParams(sort: SortRequest | undefined, gsiNames: string[]): QueryParams {
+  let queryParams: QueryParams = {};
+  if (!sort || Object.keys(sort).length === 0) {
+    return queryParams;
+  }
+  if (Object.keys(sort).length > 1) {
+    throw Boom.badRequest('Cannot sort by more than one attribute.');
+  }
+  //only one entry will exist due to validation above
+  Object.entries(sort).forEach(([key, value]) => {
```
bhudnell, line 92: Same comment as above

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: got it, I tried to allow multiple operators but found out dynamo doesn't allow it and forgot to change it back

workbench-core/base/src/interfaces/FilterRequest.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { QueryParameterFilter } from './queryParameterFilter';
+
+export interface FilterRequest {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: QueryParameterFilter<any> | undefined;
```
bhudnell, line 10: could `unknown` be used here instead of `any`?

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
bhudnell: I think this error needs to be reworded then. Maybe something like `'Cannot apply a filter and sort to different properties at the same time'`

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
bhudnell: I dont remember seeing documentation saying that only one field can be defined at a time. Probably worth adding to the interface

workbench-core/environments/src/interfaces/listEnvironmentTypeRequest.ts
FernandoAranda: changed to have plural

workbench-core/environments/src/schemas/listEnvironmentType.ts
FernandoAranda: changed to have plural

workbench-core/environments/src/schemas/listEnvironmentType.ts
FernandoAranda: changed to have plural

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: validation text updated

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: added warning to interface explaining we can only use one operator

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: changed to only retrieve first

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
FernandoAranda: changed to only retrieve first

workbench-core/base/src/interfaces/FilterRequest.ts
FernandoAranda: changed to unknown

workbench-core/base/src/interfaces/queryParameterFilter.ts
```diff
@@ -3,6 +3,11 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
+/************
+ *
+ * Only one operator can be defined by property, if multiple operators are dfined, dynamo service will throw an exception.
```
bhudnell, line 8:
```suggestion
 * Only one operator can be defined by property, if multiple operators are defined, dynamo service will throw an exception.
```

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param gsiPropertyNames - field names on pascal case format to retrieve gsi name
+ *
+ * @returns QueryParams object
+ ************************************************************/
+export function getFilterQueryParams(filter: FilterRequest | undefined, gsiNames: string[]): QueryParams {
+  let queryParams: QueryParams = {};
+  //validate filter request doesnt have more than one filter or no filters
+  if (!filter || Object.keys(filter).length === 0) {
+    return queryParams;
+  }
```
bhudnell, line 49:
```suggestion
  //validate filter request doesnt have more than one filter or no filters
  if (!filter || Object.keys(filter).length === 0) {
    return {};
  }
```

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+    const currentQueryParams: QueryParams = {
+      index: gsi,
+      sortKey: key,
+      ...filterQueryParam
+    };
+    queryParams = { ...queryParams, ...currentQueryParams };
+  } else {
+    throw Boom.badRequest('Filter contains invalid format.');
+  }
+  return queryParams;
```
bhudnell, line 75:
```suggestion
    return currentQueryParams;
  } else {
    throw Boom.badRequest('Filter contains invalid format.');
  }
```

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Creates a sort request object into a QueryParam object to be consumed by dynamoDB
+ * @param sort - sort request property object
+ * @param gsiPropertyNames - field names on pascal case format to retrieve gsi name
+ * @returns QueryParams object
+ ************************************************************/
+export function getSortQueryParams(sort: SortRequest | undefined, gsiNames: string[]): QueryParams {
+  let queryParams: QueryParams = {};
+  if (!sort || Object.keys(sort).length === 0) {
+    return queryParams;
+  }
```
bhudnell, line 88: See above suggestion

workbench-core/base/src/utilities/queryParamsFiltersUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+    const sortQueryParams = {
+      index: gsi,
+      sortKey: key,
+      forward
+    };
+    queryParams = { ...queryParams, ...sortQueryParams };
+  } else {
+    throw Boom.badRequest('Sort contains invalid format.');
+  }
+  return queryParams;
```
bhudnell, line 112: See above suggestion

bhudnell (approved):

PR #582, opened 2022-10-28
feat(pb): update LIST Projects API

maghirardelli:
Issue #, if available: GALI-1739

Description of changes:
Updated the List Projects API to support pagination and prepare for dynamic Authz integration to enforce project as a boundary.

Integ tests to come but I extensively tested the API manually through postman

I also did a small refactor to reuse pagination code and custom object types

Checklist:

### DIFF
--- a/solutions/swb-app/src/projectRoutes.ts
+++ b/solutions/swb-app/src/projectRoutes.ts
@@ -3,20 +3,44 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { ProjectService } from '@aws/workbench-core-accounts';
-import CreateProjectRequest from '@aws/workbench-core-accounts/lib/models/createProjectRequest';
+import { ProjectService, CreateProjectRequest, ListProjectsRequest } from '@aws/workbench-core-accounts';
+import Boom from '@hapi/boom';
 import { Request, Response, Router } from 'express';
 import { validate } from 'jsonschema';
 import { wrapAsync } from './errorHandlers';
 import CreateProjectSchema from './schemas/projects/createProjectSchema';
+import ListProjectsSchema from './schemas/projects/listProjectsSchema';
 import { processValidatorResult } from './validatorHelper';
 
 export function setUpProjectRoutes(router: Router, projectService: ProjectService): void {
-  // Get projects
+  // List projects
   router.get(
     '/projects',
     wrapAsync(async (req: Request, res: Response) => {
-      const projects = await projectService.listProjects();
+      let projects = {};
+      // validate if query params present
+      if (req.query.pageSize || req.query.paginationToken) {
+        const objectToValidate = {
+          pageSize: req.query.pageSize ? Number(req.query.pageSize) : undefined,
+          paginationToken: req.query.paginationToken
+        };
+        processValidatorResult(validate(objectToValidate, ListProjectsSchema));
+        // Apply pagination if applicable
+        if (
+          (objectToValidate.paginationToken && typeof objectToValidate.paginationToken !== 'string') ||
+          (objectToValidate.pageSize && Number(objectToValidate.pageSize) <= 0)
+        ) {
+          throw Boom.badRequest(
+            'Invalid pagination token and/or page size. Please try again with valid inputs.'
+          );
+        }
+        const request: ListProjectsRequest = objectToValidate as ListProjectsRequest;
+
+        projects = await projectService.listProjects(res.locals.user, request);
+      } else {
+        projects = await projectService.listProjects(res.locals.user);
+      }
+
       res.send(projects);
     })
   );
--- a/solutions/swb-app/src/projectRoutes.ts
+++ b/solutions/swb-app/src/projectRoutes.ts
@@ -4,7 +4,7 @@
  */
 
 import { ProjectService, CreateProjectRequest, ListProjectsRequest } from '@aws/workbench-core-accounts';
-import Boom from '@hapi/boom';
+import { AuthenticatedUser } from '@aws/workbench-core-authorization';
 import { Request, Response, Router } from 'express';
 import { validate } from 'jsonschema';
 import { wrapAsync } from './errorHandlers';
@@ -17,31 +17,15 @@ export function setUpProjectRoutes(router: Router, projectService: ProjectServic
   router.get(
     '/projects',
     wrapAsync(async (req: Request, res: Response) => {
-      let projects = {};
-      // validate if query params present
-      if (req.query.pageSize || req.query.paginationToken) {
-        const objectToValidate = {
-          pageSize: req.query.pageSize ? Number(req.query.pageSize) : undefined,
-          paginationToken: req.query.paginationToken
-        };
-        processValidatorResult(validate(objectToValidate, ListProjectsSchema));
-        // Apply pagination if applicable
-        if (
-          (objectToValidate.paginationToken && typeof objectToValidate.paginationToken !== 'string') ||
-          (objectToValidate.pageSize && Number(objectToValidate.pageSize) <= 0)
-        ) {
-          throw Boom.badRequest(
-            'Invalid pagination token and/or page size. Please try again with valid inputs.'
-          );
-        }
-        const request: ListProjectsRequest = objectToValidate as ListProjectsRequest;
+      const objectToValidate = {
+        user: res.locals.user as AuthenticatedUser,
+        pageSize: req.query.pageSize ? Number(req.query.pageSize) : undefined,
+        paginationToken: req.query.paginationToken
+      };
+      processValidatorResult(validate(objectToValidate, ListProjectsSchema));
+      const request: ListProjectsRequest = objectToValidate as ListProjectsRequest;
 
-        projects = await projectService.listProjects(res.locals.user, request);
-      } else {
-        projects = await projectService.listProjects(res.locals.user);
-      }
-
-      res.send(projects);
+      res.send(await projectService.listProjects(request));
     })
   );
 
--- a/solutions/swb-reference/SWBv2.postman_collection.json
+++ b/solutions/swb-reference/SWBv2.postman_collection.json
@@ -821,7 +821,7 @@
               "query": [
                 {
                   "key": "paginationToken",
-                  "value": "eyJwayI6IkVUQyIsInNrIjoiRVQjNmM3MzJlMTEtODdmYi00MGU3LWFlM2ItMjU1MTY1OGQ3OGYwRVRDIzQwYjAxNTI5LTBjN2YtNDYwOS1hMWUyLTcxNTA2OGRhNWYwZSJ9",
+                  "value": "",
                   "disabled": true
                 },
                 {
--- a/solutions/swb-reference/SWBv2.postman_collection.json
+++ b/solutions/swb-reference/SWBv2.postman_collection.json
@@ -817,7 +817,19 @@
             "url": {
               "raw": "{{API_URL}}/projects",
               "host": ["{{API_URL}}"],
-              "path": ["projects"]
+              "path": ["projects"],
+              "query": [
+                {
+                  "key": "paginationToken",
+                  "value": "eyJwayI6IkVUQyIsInNrIjoiRVQjNmM3MzJlMTEtODdmYi00MGU3LWFlM2ItMjU1MTY1OGQ3OGYwRVRDIzQwYjAxNTI5LTBjN2YtNDYwOS1hMWUyLTcxNTA2OGRhNWYwZSJ9",
+                  "disabled": true
+                },
+                {
+                  "key": "pageSize",
+                  "value": "50",
+                  "disabled": true
+                }
+              ]
             }
           },
           "response": []
--- a/workbench-core/accounts/src/models/listProjectsRequest.ts
+++ b/workbench-core/accounts/src/models/listProjectsRequest.ts
@@ -3,7 +3,10 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
+import { AuthenticatedUser } from '@aws/workbench-core-authorization';
+
 interface ListProjectsRequest {
+  user: AuthenticatedUser;
   pageSize?: number;
   paginationToken?: string;
 }
--- a/workbench-core/accounts/src/models/listProjectsRequest.ts
+++ b/workbench-core/accounts/src/models/listProjectsRequest.ts
@@ -1,0 +1,11 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+interface ListProjectsRequest {
+  pageSize?: number;
+  paginationToken?: string;
+}
+
+export default ListProjectsRequest;
--- a/workbench-core/accounts/src/services/projectService.test.ts
+++ b/workbench-core/accounts/src/services/projectService.test.ts
@@ -97,7 +97,7 @@ describe('ProjectService', () => {
       jest.spyOn(ProjectService.prototype as any, '_mockGetUserGroups').mockImplementation(() => []);
 
       // OPERATE
-      const actualResponse = await projService.listProjects(user);
+      const actualResponse = await projService.listProjects({ user });
 
       // CHECK
       expect(actualResponse).toEqual({ data: [] });
@@ -139,7 +139,7 @@ describe('ProjectService', () => {
         .resolves(queryItemResponse);
 
       // OPERATE
-      const actualResponse = await projService.listProjects(user);
+      const actualResponse = await projService.listProjects({ user });
 
       // CHECK
       expect(actualResponse.data).toEqual(items);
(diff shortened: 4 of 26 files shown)


solutions/swb-app/src/projectRoutes.ts
```diff
@@ -3,20 +3,44 @@
...
+      // validate if query params present
+      if (req.query.pageSize || req.query.paginationToken) {
+        const objectToValidate = {
+          pageSize: req.query.pageSize ? Number(req.query.pageSize) : undefined,
+          paginationToken: req.query.paginationToken
+        };
+        processValidatorResult(validate(objectToValidate, ListProjectsSchema));
+        // Apply pagination if applicable
+        if (
+          (objectToValidate.paginationToken && typeof objectToValidate.paginationToken !== 'string') ||
```
sevu1, line 30: Is it possible to confirm that schema validation will take care of types? E.g. could we try to send number and see if it's rejected by schema validation and not the code

solutions/swb-app/src/projectRoutes.ts
```diff
@@ -3,20 +3,44 @@
...
 export function setUpProjectRoutes(router: Router, projectService: ProjectService): void {
-  // Get projects
+  // List projects
   router.get(
     '/projects',
     wrapAsync(async (req: Request, res: Response) => {
-      const projects = await projectService.listProjects();
+      let projects = {};
+      // validate if query params present
+      if (req.query.pageSize || req.query.paginationToken) {
```
sevu1, line 22: nit: Invert if for readability
```
if (!req.query.pageSize && !req.query.paginationToken) {
  projects = await projectService.listProjects(res.locals.user);
  res.send(projects);
  return;
}
// rest code
```

solutions/swb-reference/SWBv2.postman_collection.json
```diff
@@ -817,7 +817,19 @@
             "url": {
               "raw": "{{API_URL}}/projects",
               "host": ["{{API_URL}}"],
-              "path": ["projects"]
+              "path": ["projects"],
+              "query": [
+                {
+                  "key": "paginationToken",
+                  "value": "eyJwayI6IkVUQyIsInNrIjoiRVQjNmM3MzJlMTEtODdmYi00MGU3LWFlM2ItMjU1MTY1OGQ3OGYwRVRDIzQwYjAxNTI5LTBjN2YtNDYwOS1hMWUyLTcxNTA2OGRhNWYwZSJ9",
```
sevu1, line 824: Do we need to provide this value in postman collection?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
    * @returns Project entries in DDB
    */
-  public async listProjects(): Promise<{ data: Project[] }> {
-    const queryParams = {
+  // TODO--add filter support
+  public async listProjects(
+    user: AuthenticatedUser,
+    request?: ListProjectsRequest
+  ): Promise<{ data: Project[]; paginationToken: string | undefined }> {
+    const pageSize = request ? request.pageSize : undefined;
```
sevu1, line 82: nit: `request?.pageSize` . Same below

sevu1: Additionally can be  ?? 0 to avoid null check later

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+          limit: pageSize && pageSize >= 0 ? pageSize : DEFAULT_API_PAGE_SIZE
+        };
+
+        queryParams = addPaginationToken(paginationToken, queryParams);
+
+        const projectsResponse = await this._aws.helpers.ddb.query(queryParams).execute();
+
+        paginationToken = getPaginationToken(projectsResponse);
+
+        return Promise.resolve({
```
sevu1, line 108: Do we need to wrap response in promise? It's already async function:

`return { data.... }`

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+    const userGroupsForCurrentUser: string[] = this._mockGetUserGroups(); // mock so the tests work
+
+    // If no group membership, return
+    if (_.isEmpty(userGroupsForCurrentUser)) {
+      return { data: [], paginationToken: undefined };
+    }
+
+    // If IT Admin, return all with pagination
+    if (userGroupsForCurrentUser.length === 1) {
+      if (userGroupsForCurrentUser[0] === 'ITAdmin') {
```
sevu1, line 95: Suggest extracting ITAdmin as const/enum

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+
+        paginationToken = getPaginationToken(projectsResponse);
+
+        return Promise.resolve({
+          data: projectsResponse.Items as unknown as Project[],
+          paginationToken: paginationToken
+        });
+      }
+      // If member of 1 group, get project item
+      if (userGroupsForCurrentUser[0] !== 'ITAdmin') {
```
sevu1, line 114: This condition is always true and can be removed

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+        const projectId = userGroupsForCurrentUser[0].split('#')[0];
+        const project = await this.getProject(projectId);
+        return Promise.resolve({ data: [project], paginationToken: undefined });
+      }
+    }
+
+    // Else, member of more than 1 group, query for project items with pagination
```
sevu1, line 121: Can ITAdmin be a member of multiple groups?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+    // welcome to party town
+    const projectIds: string[] = userGroupsForCurrentUser.map(
+      (projectGroup: string) => projectGroup.split('#')[0]
+    );
+
+    // i = 0
+    let filterExpression = 'id = :proj1';
+    const expressionAttributeValues: { [key: string]: string } = { ':proj1': projectIds[0] };
+    for (let i = 1; i < projectIds.length; i++) {
+      // build filter string "id = :proj1 OR id = :proj2 OR ... OR id = projN+1"
```
sevu1, line 131: Was `IN` operator considered?
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+      limit: pageSizeLimit,
+      filter: filterExpression,
+      values: expressionAttributeValues
     };
 
-    const projectsResponse = await this._aws.helpers.ddb.query(queryParams).execute();
+    let projects: Project[] = [];
+
+    let returnedCount = 0;
+    while (returnedCount < pageSizeLimit && returnedCount < projectIds.length) {
```
sevu1, line 151: Seems that this logic should be extracted and abstracted, something like:

`fetch<T>(params): { data: T[]; token: string }`

workbench-core/base/src/utilities/paginationHelper.ts
```diff
@@ -22,6 +22,7 @@ export function addPaginationToken(
     }
   }
 
+  console.log(params);
```
sevu1, line 25: needed?

aws-tyler (changes requested):
@maghirardelli like we discussed (and I noted in the PR), I'm going to chat with some people on the team about dynamic auth stuff. Feel free to address any of the other stuff while we figure it out. I appreciate your patience!

solutions/swb-app/src/projectRoutes.ts
maghirardelli: I'm going to include the user as part of the request per @aws-tyler 's recommendation so this check won't necessary

solutions/swb-reference/SWBv2.postman_collection.json
maghirardelli: It's just an example. Would you like me to provide a more obvious value as an example?

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Changed the code to require `request` to be defined so I think `pageSize = request.pageSize` will work because it will be `undefined` if the optional attribute of `request` does not exist.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: I will try this. I think another file was doing this so I took example from there.

solutions/swb-reference/SWBv2.postman_collection.json
sevu1: If you were to use this value would it work? I suggest using for example values only what will work, e.g. pageSize makes total sense, while for token - probably leave it blank

workbench-core/accounts/src/services/projectService.ts
maghirardelli: This is a good idea but until we know what Dynamic AuthZ provides us, I am going to keep this value as a string. I will make a note to revisit during integration with Dynamic AuthZ. That package may provide an enum of special type.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: got it--I changed it to an `else` statement

workbench-core/accounts/src/services/projectService.ts
maghirardelli: No!

workbench-core/accounts/src/services/projectService.ts
sevu1: Ha, neither `else` is needed as you have `return`. Your code looks like this:
```
if (a)
  return ;
// else is not needed because if code is here it's already an else close
// some code
```

workbench-core/accounts/src/services/projectService.ts
maghirardelli: This piece of code (getting multiple projects) will be rewritten with the new design as mentioned in meetings today so I am going to resolve this convo.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: This piece of code (getting multiple projects) will be rewritten with the new design as mentioned in meetings today so I am going to resolve this convo.

workbench-core/base/src/utilities/paginationHelper.ts
maghirardelli: nope! Thanks for catching!

workbench-core/accounts/src/services/projectService.ts
sevu1: I'd still suggest extracting enum and replace it if authz has something similar

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -83,8 +83,27 @@ describe('ProjectService', () => {
...
+      jest.spyOn(ProjectService.prototype as any, '_mockGetUserGroups').mockImplementation(() => []);
+
+      // OPERATE
+      const actualResponse = await projService.listProjects(user);
+
+      // CHECK
+      expect(actualResponse).toEqual({ data: [] });
+    });
+
+    test('list all projects as IT Admin on 1 page', async () => {
```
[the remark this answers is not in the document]
maghirardelli: pagination is built in for IT Admin case

aws-tyler:
@kpark277, @maghirardelli, and I spoke and decided server-side pagination (instead of database pagination) will have to work until we can figure out a better solution.

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -106,16 +133,431 @@ describe('ProjectService', () => {
             ':resourceType': {
               S: 'project'
             }
-          }
+          },
+          Limit: 50
```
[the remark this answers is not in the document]
maghirardelli: [Boom](https://github.com/aws-solutions/solution-spark-on-aws/blob/766becc730515dbb13095be33802ae9b0d615cba/workbench-core/base/src/utilities/paginationHelper.ts#L34)--moved from environments package so all could use it

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -106,16 +133,431 @@ describe('ProjectService', () => {
...
+
+    test('list all projects as IT Admin on more than 1 page -- getting first page', async () => {
+      // BUILD
+      const items = [projItem, projItem];
+      const lastEvaluatedKey = {
+        pk: projItem.pk,
+        sk: projItem.sk,
+        createdAt: projItem.createdAt
+      };
+      const paginationToken = Buffer.from(JSON.stringify(lastEvaluatedKey)).toString('base64');
```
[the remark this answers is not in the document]
maghirardelli: The mock won't work if we make up values here. The params generated need to match for the mock to work.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
     }
   }
 
+  // TODO--delete after dynamic Authz
+  private _mockGetUserGroups(): string[] {
```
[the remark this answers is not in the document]
maghirardelli: I believe I would still have to mock something for that to work in whatever I pass into the constructor. That was not in scope for this ticket so I am going to go with what works. This will be taken out shortly anyway.

workbench-core/accounts/src/models/listProjectsRequest.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+interface ListProjectsRequest {
+  pageSize?: number;
```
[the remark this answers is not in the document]
maghirardelli: When we set `limit`, we use a default in case nothing was passed. Plus, an interface property cannot have an initializer.

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -106,16 +133,431 @@ describe('ProjectService', () => {
...
+      // OPERATE
+      const actualResponse = await projService.listProjects(user, { pageSize });
+
+      // CHECK
+      expect(actualResponse).toEqual(expectedResponse);
+    });
+
+    test('list all projects as IT Admin on more than 1 page -- getting page with pagination token', async () => {
+      // BUILD
+      const items = [projItem, projItem];
```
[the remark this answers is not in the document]
maghirardelli: yes, after this merged, I was going to rework this test file

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+   * @param user - authenticated user listing the projects
    * @returns Project entries in DDB
    */
-  public async listProjects(): Promise<{ data: Project[] }> {
-    const queryParams = {
+  // TODO--add filter support
+  public async listProjects(
+    user: AuthenticatedUser,
+    request?: ListProjectsRequest
+  ): Promise<{ data: Project[]; paginationToken: string | undefined }> {
```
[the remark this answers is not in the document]
maghirardelli: I like this--will do.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+   * @param request - the optional fields to list projects
+   * @param user - authenticated user listing the projects
    * @returns Project entries in DDB
    */
-  public async listProjects(): Promise<{ data: Project[] }> {
-    const queryParams = {
+  // TODO--add filter support
+  public async listProjects(
+    user: AuthenticatedUser,
+    request?: ListProjectsRequest
```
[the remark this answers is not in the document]
maghirardelli: because you do not have to pass a pagtoken or pagesize but now that I put user in the request, it is required

workbench-core/accounts/src/services/projectService.ts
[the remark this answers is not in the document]
maghirardelli: resolving this comment as there will be a follow up PR to address this

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -56,20 +62,143 @@ export default class ProjectService {
...
+      if (userGroupsForCurrentUser[0] !== 'ITAdmin') {
+        const projectId = userGroupsForCurrentUser[0].split('#')[0];
+        const project = await this.getProject(projectId);
+        return Promise.resolve({ data: [project], paginationToken: undefined });
+      }
+    }
+
+    // Else, member of more than 1 group, query for project items with pagination
+    // welcome to party town
+    const projectIds: string[] = userGroupsForCurrentUser.map(
```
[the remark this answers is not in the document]
maghirardelli: We discussed--another PR is incoming to this branch to make this better

workbench-core/accounts/src/services/projectService.ts
maghirardelli: I know it is not needed, but I think it is neater to clearly have different conditionals to enter each section of code. I will get rid of it though since you don't seem to like that.

aws-tyler (approved):

PR #587, opened 2022-11-01
feat: dynamic authorization service

htztomic:
Issue #, if available:

Description of changes:

Checklist:


workbench-core/authorization/src/caslAuthorizationPlugin.test.ts
```diff
@@ -154,4 +155,94 @@ describe('CASL Authorization Plugin', () => {
...
+
+    test('Check for a invalid dynamic operations due invalid conditions', async () => {
+      const mockDynamicOperations: DynamicOperation[] = [
+        {
+          action: 'CREATE',
+          subjectType: 'SampleSubject',
+          subjectId: '1'
+        }
+      ];
+      try {
```
aws-tyler, line 217: instead of a try/catch I would use the `.toThrowError` method.

workbench-core/authorization/src/caslAuthorizationPlugin.test.ts
```diff
@@ -154,4 +155,94 @@ describe('CASL Authorization Plugin', () => {
...
+        await caslAuthorizationPlugin.isAuthorizedOnDynamicOperations(
+          mockIdentityPermissions,
+          mockDynamicOperations
+        );
+        expect.hasAssertions();
+      } catch (err) {
+        expect(err).toBeInstanceOf(ForbiddenError);
```
aws-tyler, line 224: What about using child classes for the errors to make them less generic to their context? This one could be something like `CreateForbiddenError` which inherits from `ForbiddenError`. Then we only need to check the right type is being thrown and don't need to do any matching on the message (which could change as we keep working on the project).

workbench-core/authorization/src/caslAuthorizationPlugin.ts
```diff
@@ -3,18 +3,20 @@
...
 import Permission from './permission';
+
 /**
  * {@link https://github.com/stalniy/casl | CASL } Authorization Plugin.
  *
  */
 export default class CASLAuthorizationPlugin implements AuthorizationPlugin {
   public async isAuthorized(userPermissions: Permission[], operations: Operation[]): Promise<void> {
-    const ability: Ability = this._defineAbilitiesFor(userPermissions);
+    const ability: Ability = this._defineAbilitiesForUserPermissions(userPermissions);
```
aws-tyler, line 19: Not sure if the type is declared for easy reading, but you don't need to declare it since the return type is declared on the function you're calling.

workbench-core/authorization/src/caslAuthorizationPlugin.ts
```diff
@@ -23,12 +25,36 @@ export default class CASLAuthorizationPlugin implements AuthorizationPlugin {
...
+  ): Promise<void> {
+    const ability: Ability = this._defineAbilitiesForIdentityPermissions(identityPermissions);
+    try {
+      dynamicOperations.forEach((dynamicOperation: DynamicOperation) => {
+        const attributes = {
+          subjectId: dynamicOperation.subjectId,
+          ...dynamicOperation.subjectAttributes
+        };
+        const requestedSubject = subject(dynamicOperation.subjectType, attributes);
+        CASLForbiddenError.from(ability).throwUnlessCan(
```
aws-tyler, line 41: Does CASL allow you to test the positive case i.e. `throwIfCannot` instead of needing do an `unless` statement?

workbench-core/authorization/src/caslAuthorizationPlugin.ts
```diff
@@ -38,4 +64,29 @@ export default class CASLAuthorizationPlugin implements AuthorizationPlugin {
...
+    const { can, cannot, rules } = new AbilityBuilder(Ability);
+
+    identityPermissions.forEach((identityPermission) => {
+      const conditions = {
+        ...(identityPermission.conditions ?? {}),
+        ...(identityPermission.subjectId === '*' ? {} : { subjectId: identityPermission.subjectId })
+      };
+      if (identityPermission.effect === 'ALLOW')
+        can(identityPermission.action, identityPermission.subjectType, identityPermission.fields, conditions);
+      else
```
aws-tyler, line 82: Is it possible to do an early return here so we don't need have an `else` clause?

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      jest.spyOn(mockDynamicPermissionsPlugin, 'getIdentityPermissionsBySubject').mockResolvedValueOnce({
+        identityPermissions: sampleGroupPermissions
+      });
+
+      jest.spyOn(mockDynamicPermissionsPlugin, 'getIdentityPermissionsBySubject').mockResolvedValueOnce({
+        identityPermissions: sampleGroupWildcardPermissions
+      });
+    });
+
+    test('Check user with a valid permissions for a dynamic operation', async () => {
```
aws-tyler, line 91: this test has lots of different `expects` in it. We should try to keep one expect per test to ensure we remove situations where tests end up having hidden dependencies that are needed for it to pass.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -0,0 +1,347 @@
...
+        expect.hasAssertions();
+      } catch (err) {
+        expect(err).toBeInstanceOf(ForbiddenError);
+        expect(err.message).toBe('User cannot access');
+      }
+    });
+  });
+
+  describe('createGroup', () => {
+    test('Check for dynamic permissions plugin was called', async () => {
```
aws-tyler, line 173: I'm not sure we want to be testing implementation instead of outcomes. If the implementation changes, this test will fail, so the only thing this test really tells us is "these methods were called".

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+  /**
+   * Create an authorization group
+   * @param createGroupRequest - {@link CreateGroupRequest}
+   *
+   * @returns - {@link CreateGroupResponse}
+   *
+   * @throws - {@link GroupAlreadyExistsError} Can not create a group that already exists
+   */
+  public async createGroup(createGroupRequest: CreateGroupRequest): Promise<CreateGroupResponse> {
```
aws-tyler, line 114: Why are we creating pass through methods instead of making the plugin public and accessing the plugin directly?

workbench-core/authorization/src/dynamicAuthorization/dynamicPermissionsPluginInputs.ts
```diff
@@ -0,0 +1,318 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Action } from '../action';
+import { Effect } from '../permission';
+
+/**
+ * Request object for DynamicPermissionsPlugin's createGroup
+ */
+export interface CreateGroupRequest {
```
aws-tyler, line 12: These interfaces should live in their own file to help prevent merge conflicts when two people have to make simultaneous changes to this file.

aws-tyler (commented):
I've added some comments, but this PR is too large to be able to effectively review thoroughly. 

Most of what I saw looked good! I've got a few concerns though about the tests mainly checking to see if specific methods are getting called. Those types of test are very vulnerable to mistakes, especially when the right method is called but the implementation of that method is not what you think it does. If you'd like to pair on refactoring these a bit so that the tests are robust against implementation changes just let me know!

workbench-core/base/README.md
```diff
@@ -5,7 +5,11 @@
 ## Code Coverage
 | Statements                  | Branches                | Functions                 | Lines             |
 | --------------------------- | ----------------------- | ------------------------- | ----------------- |
+<<<<<<< HEAD
+| ![Statements](https://img.shields.io/badge/statements-81.38%25-yellow.svg?style=flat) | ![Branches](https://img.shields.io/badge/branches-70.05%25-red.svg?style=flat) | ![Functions](https://img.shields.io/badge/functions-80.81%25-yellow.svg?style=flat) | ![Lines](https://img.shields.io/badge/lines-82.5%25-yellow.svg?style=flat) |
+=======
 | ![Statements](https://img.shields.io/badge/statements-81%25-yellow.svg?style=flat) | ![Branches](https://img.shields.io/badge/branches-70.25%25-red.svg?style=flat) | ![Functions](https://img.shields.io/badge/functions-79.88%25-red.svg?style=flat) | ![Lines](https://img.shields.io/badge/lines-82.11%25-yellow.svg?style=flat) |
+>>>>>>> 49e07bf68c40a14e4fc64e33aa3dd276df5bf9a7
```
bhudnell, line 12: looks like an incomplete merge here

workbench-core/authorization/src/caslAuthorizationPlugin.ts
```diff
@@ -23,12 +25,36 @@ export default class CASLAuthorizationPlugin implements AuthorizationPlugin {
...
+        };
+        const requestedSubject = subject(dynamicOperation.subjectType, attributes);
+        CASLForbiddenError.from(ability).throwUnlessCan(
+          dynamicOperation.action,
+          requestedSubject,
+          dynamicOperation.field
+        );
+      });
+    } catch (err) {
+      throw new ForbiddenError(err.message);
```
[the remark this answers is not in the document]
htztomic: I agree but I think the better option is to log it in the service

workbench-core/authorization/src/authorizationPlugin.ts
```diff
@@ -18,4 +19,14 @@ export default interface AuthorizationPlugin {
    * @throws - {@link ForbiddenError} when user is not authorized.
    */
   isAuthorized(userPermissions: Permission[], operations: Operation[]): Promise<void>;
+
+  /**
+   * Checks whether a set of {@link IdentityPermission}s is authorized to perform a set of {@link DynamicOperation}.
+   * @param identityPermissions - {@link IdentityPermission}s.
+   * @param dynamicOperations - An array  of {@link DynamicOperation}s that the user wants to perform.
```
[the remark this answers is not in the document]
htztomic: MAF has done it as a general rule, it is just more thorough documentation

workbench-core/authorization/src/caslAuthorizationPlugin.ts
htztomic: It was declared for ease of reading

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
htztomic: This could be done at the consumer's discretion, but it is rather easier to encapsulate this under one service

workbench-core/authorization/src/caslAuthorizationPlugin.ts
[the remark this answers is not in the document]
htztomic: after thinking about it more, i think this could be an anti-pattern. We are logging and throwing. This is rather inefficient because the handler could also log it again. We should just leave it to the handler to proper log it.

workbench-core/audit/src/auditLogger.ts
```diff
@@ -3,16 +3,17 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { Writer, AuditEntry } from '@aws/workbench-core-audit';
 import { LoggingService } from '@aws/workbench-core-logging';
+import AuditEntry from './auditEntry';
+import Writer from './plugins/writer';
 
-export default class AuditLogger implements Writer {
+export class AuditLogger implements Writer {
```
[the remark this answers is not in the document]
htztomic: fixed

workbench-core/authorization/src/caslAuthorizationPlugin.ts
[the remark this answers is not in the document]
aws-tyler: Is the expectation that people will be catching this error then? An example of how its supposed to work may be helpful, because it feels a little confusing that we're throwing an Auth error that the user has the option of ignoring.

workbench-core/authorization/src/authorizationPlugin.ts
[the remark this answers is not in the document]
aws-tyler: So should this have a `@throws` annotation then?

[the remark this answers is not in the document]
htztomic: good catch

workbench-core/authorization/src/caslAuthorizationPlugin.ts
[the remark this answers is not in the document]
htztomic: Yes that is the expectation. I will document more of this in a dynamicAuthz README. The idea here is we want to throw an error so regardless of what the consumer wants to do with the error, they must handle it. If they choose to handle the error by ignoring the error, the consumer has made that decision knowingly. Could jump on a call so we can talk about this further.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
aws-tyler: I think some sort of concept is missing here: the functions on the `DynamicAuthorizationService` and `DynamicAuthorizationService` are almost identical, but they're independent of each other. It looks like there is maybe an interface that both of these objects conform to at a minimum. Maybe there should be two separate classes, because it seems like we've joined Auth check functionality with group management.

workbench-core/authorization/src/caslAuthorizationPlugin.ts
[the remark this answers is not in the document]
aws-tyler: Sounds good. I'll find some time on your calendar tomorrow!

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
htztomic: What would be the benefit of separation of concerns for authz check and group management, given they both currently live under the authorization service?

workbench-core/authorization/src/caslAuthorizationPlugin.ts
htztomic: sure, is there a benefit to eliminating the else clause?

workbench-core/authorization/src/caslAuthorizationPlugin.test.ts
htztomic: I can see this as an option, but one thing to consider is that this doesn't scale well if we change `action`. We can move in the direction of using CRUD + other actions then we need errors for each action failed. I can see a scenario where consumers want to handle a specific action error but is that a valid use case right now?

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
htztomic: I agree with this

workbench-core/authorization/src/caslAuthorizationPlugin.test.ts
aws-tyler: Not a deal breaker for now, but maybe something we consider in the future to simplify error management.

workbench-core/authorization/src/caslAuthorizationPlugin.ts
aws-tyler: cyclomatic complexity management. `else` statements aren't really necessary if your code can be terminated inside an if clause.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
aws-tyler: In this specific case, we'd be improving adherence to the Single Responsibility Principle. If I understand correctly, these two functionalities are not  actually related to each other from a responsibility perspective: one checks for authorization, the other is a CRUD interface for Group management.

Are you going for a single class that the user can call to manage all things Auth related? If so, I don't think that is the direction we'd want to head because as we add more functionality the interface for that service gets more and more complex as the class takes on multiple responsibilities.

workbench-core/authorization/src/caslAuthorizationPlugin.ts
htztomic: isn't the CYC measurement here the same with or without the else clause

aws-tyler (dismissed):
Chatted with @htztomic offline and am all set on my comments!

bhudnell (approved):

workbench-core/authorization/src/caslAuthorizationPlugin.ts
aws-tyler: It is, but this is more about getting out of the practice of using else clauses that eventually would lead us down to high CYC. But like I was saying on the call, we can always refactor if it gets to that point in the future.

aws-tyler (approved):

bhudnell (approved):

PR #599, opened 2022-11-04
feat: [GALI-1966] DatabaseService.executeQuery() method

aws-tyler:
Issue #, if available:
GALI-1966

Description of changes:
Introduce an interface named `DatabaseService` that will handle codify required database functionality. 

Checklist:


workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -140,6 +142,27 @@ export default class DynamoDBService {
...
+   *
+   * @param params - optional object of optional properties to generate a query request
+   * @returns ExecuteQueryResult object
+   *
+   * @example Use this to execute a query to the DynamoDb table.
+   * ```ts
+   * const result = dynamoDBService.executeQuery({sortKey: 'value', eq: {N: '5'}});
+   * ```
+   */
+  public async executeQuery(params?: QueryParams): Promise<ExecuteQueryResult> {
```
nguyen102, line 156: What's the difference between using this method and `query.execute`? Their name sounds very similar so I'm worried developers might be confused between the two method. What are your thoughts on changing the return type of the query class to return ExecutionQueryResult?

aws-tyler: @nguyen102 we decided not to modify `query.execute` right now to prevent any breaking changes in the code. Any new code will use this new method and `query.execute` will be deprecated.

My overall plan is to remove all references to `DynamoDBService` in the code except where we're instantiating services. All the services that need to talk to dynamo will take the `DatabaseService` interface as a constructor argument, giving us a simplistic form of Dependency Injection. Any time someone needs to talk to the database, they can use the object passed in via the interface. We can add new methods as needed to `DatabaseService` and do a seamless refactor of any place `DynamoDBService` is still explicitly being used as we touch those parts of the code.

aws-tyler: Sorry, I forgot to answer what the difference is: 

Currently, `query.execute()` modifies the underlying objects in the `QueryCommandOutput`, specifically `QueryCommandOutput.Items` and `QueryCommandOutput.LastEvaluatedKey`, changing their values to `Record<string, any>` at run time, making the Typescript compiler basically useless because it can't calculate types accurately anymore. That swap also leads us to do a lot of manual type casting using `as unknown as Type`, which again effectively nullifies the typing system. The new method and gradual refactor mentioned above are meant to limit the blast radius of this issue and allow us to get back to having a reliable type checker.

nguyen102: Ah thanks for explaining. I think it makes sense to create an `DatabaseService` interface. This will help with dependency injection as you mentioned. However, I think we need to add some major changes to the interface and clearly describe how we want the interface to be used if we're deciding to go with that route for now.

`DatabaseService` interface shouldn't be tied to DDB implementation details. For example, in the current interface the parameter value is [QueryParams](https://github.com/aws-solutions/solution-spark-on-aws/blob/develop/workbench-core/base/src/constants/queryParams.ts), which is using `AttributeValue`. `AttributeValue` is a DDB type definition and ties `DatabaseService` interface with DDB types.

The interface shouldn't be DDB specific. I'm thinking the interface should look something like this

```
interface DatabaseService {
    listDBResources<T>(param: ListDBResourceRequest, expectedResourceType: T) : ListDBResourceResponse<T>
}
```

```
interface ListDBResourceResponse <T> {
   data: T[], 
   paginationToken?: string
}
```

The syntax might need some massaging, but the high level concept is calling `listDBResources` should return a list of resources that are of the same type as `expectedResourceType`. `listDBResource` should check the resources that are return from the DB and confirm that they are the wrong type.

Without this check, we're still not checking the object type and we're still in the same state as before.
This line of code `const data = result.Items?.map((item) => item as unknown as Record<string, JSONType>) || [];` is doing the same thing as what we're already doing in the other services.

maghirardelli:
A few high level things I want to discuss with you today:
1. Is having the method and returned data type too general as it is now? As we discussed, Query is just one of the DDB calls we make that has this problem of mutating the return object's type. Do you have plans for how the other DDB calls would get their versions of `executeQuery()`? I worry it might be a lot of reused code so wonder if it might be worth trying to converge on a single method for all mutations of the responses from DDB.
2. I do not think the current `executeQuery` method is either in the right service or doing the right thing exactly. The point of DDB Service is that you can create a `Query` object with this and then still have to call `.execute()` afterwards. This is so that developers can choose to hand write their `Query` objects and params (using `Query` methods) or use the DDB Service to create the params and objects (using `.query()`). Either way, the `.execute()` comes later and is triggered by the developer explicitly. I would either like to see a `DDBService.format()` method that takes in any response from DDB (kind of tying to Tim's comment above) OR a `Query.executeQuery()` method that would be used in place of `Query.execute()` (think 1:1 replacement). I don't think this method should be in `DDBService` as it is not currently a simple replacement for `.execute()`. Happy to sync online today!

workbench-core/accounts/src/services/accountService.test.ts
```diff
@@ -248,24 +250,25 @@ describe('AccountService', () => {
   test('getAccounts returns list of onboarded accounts', async () => {
     // BUILD
     const accountService = new AccountService(process.env.STACK_NAME!);
-    const accounts = [
```
aws-tyler, line 251: Our new parsing caught a test that had an invalid Account object 🎉

workbench-core/accounts/src/models/accountSchema.ts
```diff
@@ -0,0 +1,25 @@
...
+  error: ErrorSchema.optional(),
+  hostingAccountHandlerRoleArn: z.string(),
+  vpcId: z.string(),
+  subnetId: z.string(),
+  cidr: z.string(),
+  environmentInstanceFiles: z.string(),
+  encryptionKeyArn: z.string(),
+  externalId: z.string(),
+  stackName: z.string(),
+  status: z.enum(HOSTING_ACCOUNT_STATUS as [string, ...string[]])
```
aws-tyler, line 24: Because of the way `HOSTING_ACCOUNT_STATUS` is defined, its not obvious to the type checker that it matches the expected type, so we need to use an `as` to get it to both objects sufficiently overlap enough that the two types are interchangeable.

workbench-core/accounts/src/models/accountSchema.ts
```diff
@@ -0,0 +1,25 @@
...
+import { HOSTING_ACCOUNT_STATUS } from '../constants/hostingAccountStatus';
+
+// eslint-disable-next-line @rushstack/typedef-var
+const ErrorSchema = z.object({
+  type: z.string(),
+  value: z.string()
+});
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const AccountSchema = z.object({
```
nguyen102, line 16: Thoughts on naming this `accountModel` since it's in the model directory? We already have a `schema` directory, which we define the "schema" of requests. They're both similar concept, but if we use different naming convention it would help developer know the difference between a TS object and a request "schema".

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -140,6 +142,29 @@ export default class DynamoDBService {
...
+   *
+   * @param params - optional object of optional properties to generate a query request
+   * @returns Promise<PaginatedQueryResponse>
+   *
+   * @example Use this to get paginated items from the DynamoDb table.
+   * ```ts
+   * const result = dynamoDBService.getPaginatedItems({sortKey: 'value', eq: {N: '5'}});
+   * ```
+   */
+  public async getPaginatedItems(params?: QueryParams): Promise<PaginatedQueryResponse> {
```
nguyen102, line 156: Nitpick: Thoughts on renaming `PaginatedQueryResponse` to `PaginatedItemResponse`? I think the name of the method and the response type should be similar so developers can relate easily relate the two concepts. Also the `PaginatedItemResponse` doesn't have to be from a query call.

workbench-core/accounts/src/models/accountSchema.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
```
maghirardelli, line 6: why are we using zod here?

workbench-core/accounts/src/models/accountSchema.ts
```diff
@@ -0,0 +1,30 @@
...
+import { HOSTING_ACCOUNT_STATUS } from '../constants/hostingAccountStatus';
+
+// eslint-disable-next-line @rushstack/typedef-var
+const ErrorSchema = z.object({
+  type: z.string(),
+  value: z.string()
+});
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const AccountSchema = z.object({
```
maghirardelli, line 16: Why is this a `const` and not an `interface`?

maghirardelli: Why do we have this file and `account.ts`? Aren't tehy defining the same thing?

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -61,14 +62,12 @@ export default class AccountService {
...
-      accounts = response.Items.map((item: unknown) => {
-        return item as unknown as Account;
-      });
-    }
-    return accounts;
+
+    const response = await this._aws.helpers.ddb.getPaginatedItems(queryParams);
+
+    return response.data.map((item) => {
+      return AccountSchema.parse(item);
```
maghirardelli, line 69: Ah is this why we need `accountSchema`? So we can `.parse` a zod onject instead of casting `as unknown as Account` using the existing model?

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
maghirardelli: I would also prefer this method to explicitly mention query since it is the wrapper for `query()`. Why did you change it from `executeQuery()`?

workbench-core/accounts/src/models/accountSchema.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from 'zod';
+import { HOSTING_ACCOUNT_STATUS } from '../constants/hostingAccountStatus';
+
+// eslint-disable-next-line @rushstack/typedef-var
```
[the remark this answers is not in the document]
aws-tyler: rush wants us declare a type for any variable we instantiate instead of having it be calculated by the type system. The Zod types are that get returned are extremely verbose and it would make this code harder to read.

workbench-core/accounts/src/models/accountSchema.ts
aws-tyler: The object getting created here is a Zod runtime parser, not an `Account`. We can call it a schema if we want to, but our existing schemas are really only used for validation right now. What about putting it in `src/models/parsers` since these parsers belong to the models?

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
aws-tyler: @maghirardelli I felt like I could better describe what the method was doing with this name over `executeQuery`, which doesn't give any indication that the you're potentially getting back more than one item and that pagination could be involved.

workbench-core/accounts/src/models/accountSchema.ts
aws-tyler: For runtime and transpile time type checking and data validation.

workbench-core/accounts/src/models/accountSchema.ts
aws-tyler: `AccountSchema` isn't an `interface`, it is a Zod parser that parses and validates our data into the `Account` interface format.

workbench-core/accounts/src/models/accountSchema.ts
nguyen102: Oh if it's a parser, should we just call it `AccountParser` and `ErrorParser`. Should we then put this code in the file `Account.ts`? The interface there is highly coupled with the parser, so if a change is made to the interface, we'll know to update the parser as well.

workbench-core/accounts/src/services/accountService.ts
aws-tyler: We use the Zod parser to parse the `item` that comes back from `getPaginatedItems`. It validates AND parses the data for us, guaranteeing that TSC is getting the static type correct and that our transpiled Javascript code is getting the types we expect at runtime.

workbench-core/accounts/src/models/account.ts
```diff
@@ -5,7 +5,7 @@
 
 import { HostingAccountStatus } from '../constants/hostingAccountStatus';
 
-interface Account {
+export default interface Account {
```
nguyen102, line 8: Btw, instead of using an interface. Could we just use a `type`? 
```
export type Account =  z.infer<typeof AccountParser>;
```

workbench-core/accounts/src/models/accountSchema.ts
aws-tyler: Good idea. I like that. I wasn't sure about including everything in a single file (I'm used to putting everything in its own in other projects), so didn't really consider that as a viable solution, but I definitely think it makes sense here.

workbench-core/accounts/src/models/account.ts
nguyen102: This way we don't have to define an account twice. Once in the parser and the second time in the interface.

aws-tyler: I can but then everything is defined in terms of Zod types, which I'm not sure we want to do. The repeated declarations is a little annoying but it saves us readability 🤷 . Thoughts?

Screen Shot 2022-11-08 at 12 53 57 PM

aws-tyler: When you actually instantiate the type, i believe that get cleaned up and you just see `string` etc, but not when you're defining the `type`

aws-tyler: Actually it doesn't 😢. I'd suggest keeping both declarations for now so we don't have to use Zod types everywhere. Its an unfortunate trade off but that's what happens when static type analysis is built in as an afterthought at compile time. 
Screen Shot 2022-11-08 at 12 56 43 PM

nguyen102: Ah I see. In that case I'm ok with defining Account twice.

nguyen102 (dismissed):

common/changes/@aws/workbench-core-base/atmikev-GALI-1966_ExecuteQueryMethod_2022-11-08-17-42.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@aws/workbench-core-base",
+      "comment": "Fixed mistaken type",
```
bhudnell, line 5: Can we be more descriptive here?

aws-tyler: Updated!

bhudnell (approved):

nguyen102 (approved):

FernandoAranda (approved):

PR #613, opened 2022-11-10
feat: [GALI-1785] list hosting accounts

aws-tyler:
Issue #, if available:

Description of changes:
Added support for `GET /aws-accounts`

Checklist:


workbench-core/accounts/src/models/accounts/account.ts
```diff
@@ -46,5 +28,10 @@ export const AccountParser = z.object({
...
   externalId: z.string(),
   stackName: z.string(),
-  status: z.enum(HOSTING_ACCOUNT_STATUS as [string, ...string[]])
+  status: z.enum(HOSTING_ACCOUNT_STATUS as [string, ...string[]]),
+  updatedAt: z.string(),
+  createdAt: z.string(),
+  costCenter: CostCenterParser.optional()
 });
+
+export type Account = z.infer<typeof AccountParser>;
```
FernandoAranda, line 37: I really like this

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -19,11 +19,19 @@ import { escape } from 'lodash';
 import { wrapAsync } from './errorHandlers';
 import { processValidatorResult } from './validatorHelper';
 
-export function setUpAccountRoutes(router: Router, account: HostingAccountService): void {
+export function setUpAccountRoutes(router: Router, hostingAccountService: HostingAccountService): void {
+  router.get(
+    '/aws-accounts',
+    wrapAsync(async (req: Request, res: Response) => {
+      const listRequest = ListAccountsRequestParser.parse(req.params);
```
FernandoAranda, line 26: I think we should throw a bad request instead of a 500, or  if we don't want it to be part of the scope we should add a ticket so we don't forget it in the future

workbench-core/accounts/src/models/accounts/listAcountsRequest.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const ListAccountsRequestParser = z.object({
```
nguyen102, line 4: Can we add a test for this? It should throw `Boom.badInput` exceptions, and the message should be customer friendly.

solutions/swb-app/src/accountRoutes.ts
nguyen102: Interesting, are we moving away from using json-schema to validate requests? That's what is being used in other APIs. For example, line 41 of this file uses schema.

We might want to differentiate between the schema of the API params and the Request object for the service itself. Doing it this way means that `req.params` have to have the same attribute as the `ListAccountRequest` object. Is this a restriction we're ok with imposing on all APIs?

For example, what if we wanted to add an extra attribute to `listRequest` that is not a part of `req.params`? Like for example, we want to add a feature flag to `listRequest` before sending the request to `hostingAccountService.list`.

Note, we also have plans to move request schemas to `swb-app`. [(link)](https://quip-amazon.com/SmkEAs1xqZBy/SWBv2-Request-Object-and-Schema)

This allows developer to choose their own schema for the API. And then it's up to the developer to convert the schema to the correct request object for using the `ResourceService`.

FernandoAranda: How I see it is that the same way json schema only validates the request object , zod will only validate the request object, and then we are sure the object has the format we need so we can just parse the request into the service object if its needed, in this case there was nothing needed but at least the service doesn't have to worry about the format of the object, it will only validate values not formats

solutions/swb-reference/SWBv2.postman_collection.json
```diff
@@ -817,7 +817,19 @@
             "url": {
               "raw": "{{API_URL}}/projects",
               "host": ["{{API_URL}}"],
-              "path": ["projects"]
+              "path": ["projects"],
```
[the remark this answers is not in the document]
aws-tyler: Thought I caught them all but missed this one i guess 🙏

workbench-core/accounts/src/models/accounts/listAcountsRequest.ts
aws-tyler: We don't want to throw the exception in here right? We want Zod to give us back all the errors together so that we can combine them in the error response, that way a user can fix everything at once and not one per response. The test needs to be at the controller/router level (which is the place that should be throwing the error) since this is a validation check. Its my understanding that we do controller/router level tests at the integration level, so I was leaving those for that.

nguyen102: Ah got it. Let's get all the errors and provide one message for the requester.

solutions/swb-app/src/generateRouter.ts
```diff
@@ -104,7 +103,6 @@ export function generateRouter(apiRouteConfig: ApiRouteConfig): Express {
   setUpUserRoutes(router, apiRouteConfig.userManagementService);
   setUpEnvTypeRoutes(router, apiRouteConfig.environmentTypeService);
   setUpEnvTypeConfigRoutes(router, apiRouteConfig.environmentTypeConfigService);
-  setUpProjectRoutes(router, apiRouteConfig.projectService);
```
kpark277, line 107: Was this intentional?

aws-tyler: I pulled in some of Marianna's code that added projects functionality. Removed all those references but if this existed before then I'm going to have to do some sort of diff between my feature branch and develop :sigh:

solutions/swb-app/src/accountRoutes.ts
kpark277: I'm in favor of moving to zod to get additional functionality, while still achieving the same goals that json schema achieves. The goal of a validator for requests is to ensure that all req.params match the expected request object anyways, so I don't see the need for handling extra attributes that aren't clearly defined in the request parser.

nguyen102: I'm ok with using zod for validating api requests and service request. However, the api request schema and service request schema shouldn't be tightly coupled. Another solution might want to use our services but might want different values for the api request parameters. For the example we might use project as part of our api param but DEA might want to use group as the api param instead.

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -17,13 +20,21 @@ import { Request, Response, Router } from 'express';
...
 import { escape } from 'lodash';
 import { wrapAsync } from './errorHandlers';
-import { processValidatorResult } from './validatorHelper';
+import { processValidatorResult, validateAndParse } from './validatorHelper';
+
+export function setUpAccountRoutes(router: Router, hostingAccountService: HostingAccountService): void {
+  router.get(
+    '/aws-accounts',
+    wrapAsync(async (req: Request, res: Response) => {
+      const validatedRequest = validateAndParse<ListAccountRequest>(ListAccountsRequestParser, req.query);
```
nguyen102, line 29: IMO: This sets a bad example for other solutions that wants to use our services because their API would have to have the same field as the service request. Other solutions might want to use different name for the fields in their API. For example, instead of the API request being `{ projects:['proj1']}` the API might be `{ groups: ['group1']}`. This example dictate that the service field must match the API field. This work well for us because we're writing the service and the API.

There should be a `ListAccountsApiRequestParser` that lives in `swb-app/apiModel/account` that describe the shape of the request. In this case it could have the same Zod type definition as  `ListAccountsApiRequestParser` since in our use case the two request are the same. 

We should follow that as an example for future APIs. We shouldn't tightly couple the API request with the service request. 

If we follow this format, DEA can use this as an example for their package `dea-app`. 

An example where they talk about it is [here](https://stackoverflow.com/a/53981773)

I think the code should look something like this
```
// Throw 400 error if request is not valid
validate(ListAccountsApiRequestParser, req.query);
// Map the API Request to the Service Request
const listAccountRequest = getServiceRequest<ListAccountRequest>(ListAccountsRequestParser, req.query, listAccountApiToServiceRequestMapper);

res.send(await hostingAccountService.list(listAccountRequest));
```

With that being said, I'm not going to block the PR on this change. When DEA or another solution wants to use a services in which the service request definition and API definition is different, they can adopt the code example above.

workbench-core/accounts/src/services/accountService.test.ts
```diff
@@ -272,59 +247,119 @@ describe('AccountService', () => {
...
+      const actualResponse = await accountService.getAccounts({});
+      expect(actualResponse).toEqual({
+        data: [AccountParser.parse(accountJson)],
+        paginationToken
+      });
+    });
+
+    describe('when there is no pagination token from dynamo', () => {
+      beforeEach(() => {
+        paginationToken = undefined;
```
nguyen102, line 316: We don't need a`beforeEach` for this test right? There's only one test so `paginationToken` can just go on line 320.

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -33,21 +39,30 @@ export default class AccountService {
...
-   * Get all account entries from DDB
+   * Get account entries from DDB
    *
    * @returns Account entries in DDB
    */
-  public async getAccounts(): Promise<Account[]> {
-    const queryParams = {
-      index: 'getResourceByCreatedAt',
-      key: { name: 'resourceType', value: 'account' }
+  public async getAccounts(queryParams: QueryParams): Promise<PaginatedResponse<Account>> {
```
nguyen102, line 46: Thoughts on naming this `getPaginatedAccounts`.

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -33,21 +39,30 @@ export default class AccountService {
...
+      paginationToken: response.paginationToken
     };
+  }
 
-    const response = await this._aws.helpers.ddb.getPaginatedItems(queryParams);
-
-    return response.data.map((item) => {
-      return AccountParser.parse(item);
-    });
+  public async getAccountsForAccountHandler(queryParams: QueryParams): Promise<Account[]> {
```
nguyen102, line 57: Thoughts on name this `getAllAccounts`. We should name the function by what it does instead of by who is currently using it. This function doesn't have to be exclusively used by `AccountHandler`.

workbench-core/accounts/src/services/costCenterService.test.ts
```diff
@@ -118,36 +121,41 @@ describe('CostCenterService', () => {
...
       describe('`dependency` is the id of a saved Account', () => {
         beforeEach(() => {
           ddbMock.on(GetItemCommand).resolves({
-            Item: marshall(accountMetadata, {
+            Item: marshall(account, {
               removeUndefinedValues: true
             })
           });
-
+          accountId = 'whatever';
```
nguyen102, line 143: lol, interesting choice. Why don't we just keep what it originally was?
`${resourceTypeToKey.account.toLowerCase()}-sampleAccId`;`

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -4,17 +4,23 @@
...
+import { CostCenterParser } from '../models/costCenter/costCenter';
 
 export default class AccountService {
-  private _aws: AwsService;
+  private readonly _dynamoDBService: DynamoDBService;
 
-  public constructor(tableName: string) {
-    this._aws = new AwsService({ region: process.env.AWS_REGION!, ddbTableName: tableName });
+  public constructor(dynamoDBService: DynamoDBService) {
+    this._dynamoDBService = dynamoDBService;
```
nguyen102, line 23: Nice change!

workbench-core/accounts/src/services/accountService.test.ts
aws-tyler: The `beforeEach` function is a place to set up the described scenario for the code. In this case, we're setting up a scenario where there is no pagination token. Using a before each means that we don't have to copy and past that set up for any other tests in this scenario since they would all need `pagionationToken` to equal `undefined`.

solutions/swb-app/src/validatorHelper.ts
```diff
@@ -21,4 +22,22 @@ function processValidatorResult(validatorResult: ValidatorResult): void {
...
-export { processValidatorResult };
+function validateAndParse<T>(parser: ZodTypeAny, data: unknown): T {
+  const parsed = parser.safeParse(data);
+
+  if (parsed.success) {
+    return parsed.data;
+  }
+
+  const errorMessages = parsed.error.issues.map((issue) => {
+    return `${issue.path.join('.')}: ${issue.message}`;
```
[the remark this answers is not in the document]
aws-tyler: Examples are available [here](https://github.com/colinhacks/zod/blob/master/ERROR_HANDLING.md#zodissue).

workbench-core/accounts/src/models/accounts/listAccountsRequest.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    describe('is less than 1', () => {
+      beforeEach(() => {
+        requestObject = {
+          pageSize: '0'
+        };
+      });
+
+      test('it throws an error', () => {
+        const parsed = ListAccountsRequestParser.safeParse(requestObject);
+        expect(parsed.success).toEqual(false);
```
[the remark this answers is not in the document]
aws-tyler: I forgot to update the name of the test, which may be where the confusion is. But if you're still expecting that this would throw a Boom message, no that should not be thrown here. We're using `safeParse`, which returns a parsed result that either has the data if the parsing was successful, or a set of issues if the parsing was a failure. If you're asking why we aren't validating the error message coming back on the issue, I didn't add that but will update it to be present.

workbench-core/accounts/src/services/accountService.test.ts
nguyen102: Ah got it. Thanks for the explanation.

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -3,12 +3,15 @@
...
-// AWS Account management
 import {
   CreateAccountSchema,
   UpdateAccountSchema,
   HostingAccountService
 } from '@aws/workbench-core-accounts';
+import {
+  ListAccountRequest,
+  ListAccountsRequestParser
+} from '@aws/workbench-core-accounts/lib/models/accounts/listAccountsRequest';
```
bhudnell, line 14: should these be added to the index file of `workbench-core/accounts` so the import isnt as clunky?

workbench-core/accounts/src/services/accountService.test.ts
```diff
@@ -15,19 +15,33 @@ import {
 import { ServiceInputTypes, ServiceOutputTypes } from '@aws-sdk/client-s3';
 import { marshall } from '@aws-sdk/util-dynamodb';
 import { resourceTypeToKey } from '@aws/workbench-core-base';
+import DynamoDBService from '@aws/workbench-core-base/lib/aws/helpers/dynamoDB/dynamoDBService';
+import JSONValue from '@aws/workbench-core-base/lib/types/json';
```
bhudnell, line 19: should these be included in the index file as well?

workbench-core/accounts/src/services/accountService.ts
```diff
@@ -4,17 +4,23 @@
...
 
 import { GetItemCommandOutput } from '@aws-sdk/client-dynamodb';
-import { AwsService, resourceTypeToKey, uuidWithLowercasePrefix } from '@aws/workbench-core-base';
+import {
+  PaginatedResponse,
+  QueryParams,
+  resourceTypeToKey,
+  uuidWithLowercasePrefix
+} from '@aws/workbench-core-base';
+import DynamoDBService from '@aws/workbench-core-base/lib/aws/helpers/dynamoDB/dynamoDBService';
```
bhudnell, line 13: include in index file?

bhudnell (dismissed):

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -47,7 +54,17 @@ export default class HostingAccountLifecycleService {
...
+  }
+
+  public async listAccounts(listAccountRequest: ListAccountRequest): Promise<PaginatedResponse<Account>> {
+    const queryParams = addPaginationToken(listAccountRequest.paginationToken, {
+      index: 'getResourceByName',
+      key: { name: 'resourceType', value: 'account' },
+      limit: listAccountRequest.pageSize || DEFAULT_API_PAGE_SIZE
+    });
+
+    return await this._accountService.getPaginatedAccounts(queryParams);
```
nguyen102, line 67: I don't think you need to `return await ....`, you can just return the promise as `return ...`

nguyen102 (approved):
Looks good. Couple of small comments.

workbench-core/accounts/src/models/accounts/listAccountsRequest.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        requestObject = {
+          pageSize: '101'
+        };
+      });
+
+      test('it returns an error', () => {
+        const parsed = ListAccountsRequestParser.safeParse(requestObject);
+        expect(parsed.success).toEqual(false);
+
+        if (!parsed.success) {
```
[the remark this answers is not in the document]
aws-tyler: The TSC requires you do these checks when the type is a [Union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#working-with-union-types)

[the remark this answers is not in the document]
nguyen102: thanks for the explanation.

nguyen102 (approved):

bhudnell (approved):

workbench-core/example/infrastructure/src/__tests__/__snapshots__/example-stack.test.ts.snap
```diff
@@ -252,7 +252,7 @@ Object {
         },
         "FunctionVersion": Object {
           "Fn::GetAtt": Array [
-            "ExampleLambdaServiceCurrentVersionDB4EDA2B9b9c79fb44cad026e6114dcbb2adeec6",
```
insignias, line 255: non-blocking: snapshot test has been deleted for example App in the latest merge to develop. So this file should not be present

insignias (approved):

PR #620, opened 2022-11-11
feat: filter and sort on list projects API

maghirardelli:
Issue #, if available: GALI-1958/GALI-1959

Description of changes:
Adds new implementation for listing projects when a user is part of many project groups
Add filtering and sorting support for the List Projects API
Small improvements to pagination and query utilities

Checklist:


workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
       return { data: [project], paginationToken: undefined };
     }
 
-    // Else, member of more than 1 group, query for project items with pagination
-    // welcome to party town
```
bhudnell, line 131: sad to see this comment go 😢

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
-      if (projectsResponse.Items === undefined || _.isEmpty(projectsResponse.Items)) {
-        paginationToken = getPaginationToken(projectsResponse);
-        console.log(paginationToken);
-      }
-
-      // nothing else to get from DDB
-      else if (projectsResponse.LastEvaluatedKey === undefined) {
-        projects = projects.concat(projectsResponse.Items.map((item) => this._formatFromDDB(item)));
-        paginationToken = undefined;
+    const keys: { [key: string]: unknown }[] = projectIds.map((projectId) =>
```
bhudnell, line 148:
```suggestion
    const keys: Record<string, unknown>[] = projectIds.map((projectId) =>
```

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
+        projectsOnPage = manualFilterProjects(filter, projects);
       }
-
-      // too little--while loop continues
-      else if (returnedCount < pageSize) {
-        projects = projects.concat(projectsResponse.Items.map((item) => this._formatFromDDB(item)));
-        paginationToken = getPaginationToken(projectsResponse);
+      if (sort) {
+        projectsOnPage = manualSortProjects(sort, projects);
+      } else if (filter === undefined) {
```
bhudnell, line 163: should this be `if (sort === undefined)`?

bhudnell: if so, can be replaced with just an `else`

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
-        projects = projects.concat(projectsResponse.Items.map((item) => this._formatFromDDB(item)));
+      // build page and pagination token
+      if (paginationToken) {
+        const manualExclusiveStartKey = fromPaginationToken(paginationToken);
+        const exclusiveStartProjectId = manualExclusiveStartKey.pk.split('#')[1];
+        const exclusiveStartProject = projectsOnPage.find(
+          (project) => project.id === exclusiveStartProjectId
+        );
+        if (exclusiveStartProject === undefined) {
+          throw Boom.badImplementation('Something went wrong with listing projects');
```
bhudnell, line 175: should this be more descriptive?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
+        projectsOnPage = projectsOnPage.slice(0, pageSize);
+        const manualLastEvaluatedKey = buildDynamoDBPkSk(
+          projectsOnPage[pageSize - 1].id,
+          resourceTypeToKey.project
+        );
+        return { data: projectsOnPage, paginationToken: toPaginationToken(manualLastEvaluatedKey) };
       }
+    } else {
+      return { data: [], paginationToken: undefined };
     }
```
bhudnell, line 195: It was a little confusing getting here and going back to see what the if was for this else. Might improve readability if we early return if `projectsResponse.Responses` is falsy

workbench-core/base/src/utilities/paginationHelper.ts
```diff
@@ -31,4 +31,24 @@ export function getPaginationToken(ddbQueryResponse: QueryCommandOutput): string
     : undefined;
 }
 
+/**
+ * Function to convert an object of key-value string pairs to a pagination token.
+ *
+ * @param key - object of key-value pairs to convert to a pagination token
+ * @returns string of a pagination token
+ */
+export function toPaginationToken(key: { [key: string]: string }): string {
```
bhudnell, line 40:
```suggestion
export function toPaginationToken(key: Record<string, string>): string {
```

workbench-core/base/src/utilities/paginationHelper.ts
```diff
@@ -31,4 +31,24 @@ export function getPaginationToken(ddbQueryResponse: QueryCommandOutput): string
...
+  return Buffer.from(JSON.stringify(key)).toString('base64');
+}
+
+/**
+ * Function to convert a pagination token to an object of key-value string pairs.
+ *
+ * @param token - pagination token to convert to an object of key-value string pairs
+ * @returns object of key-value string pairs
+ */
+export function fromPaginationToken(token: string): { [key: string]: string } {
```
bhudnell, line 50:
```suggestion
export function fromPaginationToken(token: string): Record<string, string> {
```

workbench-core/accounts/src/services/projectService.ts
maghirardelli: me too, but the rewrite isn't quite worthy of party town status

workbench-core/accounts/src/services/projectService.ts
maghirardelli: this third condition is if filter and sort are both not defined. We need to handle the cases:
1. filter is defined --> apply filter
2. sort is defined --> apply sort
3. filter and sort are defined --> apply filter and then sort (there are checks above that verify the attribute to filter and sort on are the same) so no `else if` between the first two checks
4. filter and sort are undefined --> apply default sort.

So it cannot just be an `else` because filter might have been defined above. The `else` part of the `else if` will be true when sort is undefined and the `if` condition of this checks to make sure filter was also undefined.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: This check would fail if the pagination token is invalid, if the pk is not formatted right from the parsed pagination token, or if the id from the pagination token doesn't exist in the list of projects. Do you have a suggestion for the error message?

workbench-core/accounts/src/services/projectService.ts
maghirardelli: so switch the `if` condition in line 152 to be a `!projectsResponse.Responses`? How would you feel about a comment in the else block says `// no responses returned from DDB`. I feel like having the `if` being the negative case isn't normal.

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -415,189 +500,808 @@ describe('ProjectService', () => {
...
+      const user: AuthenticatedUser = {
+        id: 'user-123',
+        roles: ['ITAdmin']
+      };
+
+      // mock getUserGroups--TODO update after dynamic AuthZ intergration
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      jest.spyOn(ProjectService.prototype as any, '_mockGetUserGroups').mockImplementation(() => ['ITAdmin']);
+
+      ddbMock
```
aws-tyler, line 686: In the future we should probably not mock the database and instead mock at the service layer. In this case we'd be mocking the `execute()` method on Query. That helps us decouple from the underlying library we use to connect to dynamo. As its currently written, we have to go and replace all the tests if we swap libraries even though the tests don't really care about what is being used to call dynamo.

workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -415,189 +500,808 @@ describe('ProjectService', () => {
...
+      const actualResponse = await projService.listProjects({
+        user,
+        sort: { name: 'asc' }
+      });
+
+      // CHECK
+      expect(actualResponse.data).toEqual([proj]);
+    });
+
+    test('list all projects as IT Admin on more than 1 page -- getting first page', async () => {
```
aws-tyler, line 658: There is a lot of copy and past here. How could we DRY up these tests?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -85,10 +92,13 @@ export default class ProjectService {
    * @param request - the request object for listing projects
    * @returns Project entries in DDB
    */
-  // TODO--add filter support
   public async listProjects(request: ListProjectsRequest): Promise<ListProjectsResponse> {
-    const pageSize = request.pageSize && request.pageSize >= 0 ? request.pageSize : DEFAULT_API_PAGE_SIZE;
-    let paginationToken = request.paginationToken;
+    // Get the values from request
+    const { filter, sort } = request;
+    let { pageSize, paginationToken } = request;
```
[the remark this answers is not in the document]
maghirardelli: I grouped extracting the information from the `request` at the top here. All of these values might be used starting in line 112 for the IT Admin case.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -85,10 +92,13 @@ export default class ProjectService {
...
+    pageSize = pageSize && pageSize >= 0 ? pageSize : DEFAULT_API_PAGE_SIZE;
```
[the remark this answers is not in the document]
maghirardelli: Good call out. This is something we have done in other files but I agree it should not happen. I will add a check to error on `<= 0`

workbench-core/accounts/src/services/projectService.ts
bhudnell: I think we should remove the `else if` and breaking that login into its own conditional would be best for readability

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -107,6 +117,10 @@ export default class ProjectService {
           limit: pageSize
         };
 
+        const filterQuery = getFilterQueryParams(filter, listProjectGSINames);
+        const sortQuery = getSortQueryParams(sort, listProjectGSINames);
+        queryParams = { ...queryParams, ...filterQuery, ...sortQuery };
```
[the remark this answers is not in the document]
maghirardelli: I thought this was more readable to be honest.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Yeah, that works as well and probably does increase readability. Will add!

workbench-core/accounts/src/services/projectService.ts
maghirardelli: I do not have a suggestion for a more descriptive error message considering the conditions this would fail. IMO, when this fails, it is likely not something the user intentionally did and I have no suggestion or way to pinpoint exactly what went wrong. Hence, says something went wrong. That is why I asked for suggestions and explained the conditions in case anyone else had a better idea!

bhudnell: It seems like all 3 options are related to an invalid pagination token. Maybe something like `Invalid pagination token`?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
-        paginationToken = undefined;
+    const keys: { [key: string]: unknown }[] = projectIds.map((projectId) =>
+      buildDynamoDBPkSk(projectId, resourceTypeToKey.project)
+    );
+    const projectsResponse = (await this._aws.helpers.ddb.get(keys).execute()) as BatchGetItemCommandOutput;
+    if (projectsResponse.Responses) {
+      const projects: Project[] = projectsResponse.Responses[this._tableName].map((item) =>
+        this._formatFromDDB(item)
+      );
+      // apply sort or filter
```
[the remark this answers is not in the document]
maghirardelli: If I made this a separate function, it would be private so then we could not unit test it. Since this is just one line essentially, I am going to keep it like this.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
+      } else if (filter === undefined) {
+        // default sort is by createdAt
+        projectsOnPage = manualSortProjects({ createdAt: 'asc' }, projects);
       }
-
-      // juuuust the right amount--getPaginationToken from response and exit loop
-      else if (returnedCount === pageSize) {
-        paginationToken = getPaginationToken(projectsResponse);
-        projects = projects.concat(projectsResponse.Items.map((item) => this._formatFromDDB(item)));
+      // build page and pagination token
```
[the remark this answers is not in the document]
maghirardelli: I can try this! I wanted to write it in the method at first so I could easily work through the implementation like I didn't know exactly what the params would need to be passed to the helper method, but now that I know, I can try to extract it.

workbench-core/accounts/src/services/projectService.ts
bhudnell: I agree with Tyler. If it helps, the way I think of early returns is "if this case happens, dont worry about continuing on, end here"

workbench-core/accounts/src/services/projectService.ts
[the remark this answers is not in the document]
maghirardelli: So, do you think `badRequest` is better? The error could be due to an invalid pagination token or just something going wrong in finding the id from the pagination token, so I didn't know which one was best.

workbench-core/accounts/src/services/projectService.ts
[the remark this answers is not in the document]
aws-tyler: You wouldn't make it a private method, you'd make it an object whose responsibility is filtering and sorting. The methods would be public on that object.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
+      } else {
+        projectsOnPage = projectsOnPage.slice(0, pageSize);
+        const manualLastEvaluatedKey = buildDynamoDBPkSk(
+          projectsOnPage[pageSize - 1].id,
+          resourceTypeToKey.project
+        );
+        return { data: projectsOnPage, paginationToken: toPaginationToken(manualLastEvaluatedKey) };
       }
+    } else {
+      return { data: [], paginationToken: undefined };
```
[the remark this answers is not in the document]
maghirardelli: Yup, I swapped the `if` and `else` conditions

workbench-core/accounts/src/services/projectService.ts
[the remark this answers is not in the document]
aws-tyler: Because the pagination token is being sent in by the user, it should be a 4xx: probably a 422 since it is syntactically correct but logically unprocessable, but it looks like we've been going with 400s for validation errors.

workbench-core/accounts/src/utilities/projectUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export function manualSortProjects(sort: ProjectSort, projects: Project[]): Project[] {
+  // if no sort attribute, return original list
+  if (!sort || Object.keys(sort).length === 0) {
+    return projects;
+  }
+
+  // if too many sort attributes, throw error
+  if (Object.keys(sort).length > 1) {
+    throw Boom.badRequest('Cannot sort by more than one attribute.');
```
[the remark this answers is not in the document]
maghirardelli: I am not sure. I thought we used `Boom` everywhere in the code bases. For example, other utils throw `Boom` errors so I wanted to keep it consistent.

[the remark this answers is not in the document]
aws-tyler: I would be pretty confused as a developer if I tried to use a util function outside of a web server context and got an HTTP error back, but that may just be me.

workbench-core/accounts/src/utilities/projectUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+      return projectValue >= selectedValue;
+    case 'between':
+      if (!(selectedValue.value1 && selectedValue.value2)) {
+        throw Error('Need two values for between operation');
+      }
+      return projectValue >= selectedValue.value1 && projectValue <= selectedValue.value2;
+    case 'begins':
+      return selectedValue.startsWith(selectedValue);
+    default:
+      throw Error('invalid comparison');
```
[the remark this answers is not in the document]
maghirardelli: Good catch. This was an oversight as you included `Error`s in the code snippet you sent me and I honestly didn't realize they were not `Boom`s. Will switch to use `Boom`. Thanks!

workbench-core/accounts/src/utilities/projectUtils.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      }
+    ];
+  });
+
+  describe('manualSortProjects', () => {
+    let sortRequest: ProjectSort;
+    let expectedProjects: Project[];
+
+    test('should return original list if no sort passed in', async () => {
+      expect(manualSortProjects(sortRequest, projects)).toEqual(projects);
```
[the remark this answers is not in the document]
maghirardelli: Yup, as we have discussed in other PRs to this branch, I would like your help refactoring these tests before the feature branch is merged to develop. I am focusing on moving fast right now, and this is how I know how to write test cases. I definitely want your insight/pairing if you're open to it to help me with the refactor because I haven't written really DRY unit tests before. I have been trying to reduce duplication where I can (even these tests are better than past suites IMO), but I need a little more guidance and wanted to tackle them all at once.

workbench-core/accounts/src/utilities/projectUtils.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      // TEST n CHECK
+      expect(() => manualSortProjects(sortRequest, projects)).toThrow(
+        `No notAnAttribute on Project. Please sort by another attribute.`
+      );
+    });
+
+    test('sort on createdAt ascending', async () => {
+      // BUILD
+      sortRequest = { createdAt: 'asc' };
+      expectedProjects = projects;
```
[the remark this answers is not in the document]
maghirardelli: I am not sure what you mean by this comment. Can you elaborate?

workbench-core/accounts/src/utilities/projectUtils.test.ts
```diff
@@ -0,0 +1,526 @@
...
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ProjectStatus } from '../constants/projectStatus';
+import { ProjectFilter, ProjectSort } from '../models/listProjectsRequest';
+import Project from '../models/project';
+import { manualFilterProjects, manualSortProjects } from './projectUtils';
+
+describe('projectUtils', () => {
```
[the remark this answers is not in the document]
maghirardelli: See above comment about refactoring unit tests with your help for the entire branch after feature work is merged in!

workbench-core/accounts/src/services/projectService.test.ts
maghirardelli: See above comment about refactoring unit tests with your help for the entire branch after feature work is merged in!

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -127,91 +141,58 @@ export default class ProjectService {
...
-
-      // nothing else to get from DDB
-      else if (projectsResponse.LastEvaluatedKey === undefined) {
-        projects = projects.concat(projectsResponse.Items.map((item) => this._formatFromDDB(item)));
-        paginationToken = undefined;
+    const keys: { [key: string]: unknown }[] = projectIds.map((projectId) =>
+      buildDynamoDBPkSk(projectId, resourceTypeToKey.project)
+    );
+    const projectsResponse = (await this._aws.helpers.ddb.get(keys).execute()) as BatchGetItemCommandOutput;
+    if (projectsResponse.Responses) {
```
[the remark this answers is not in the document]
maghirardelli: Yup, I swapped the cases following Brendon and Tyler's advice

workbench-core/base/src/utilities/paginationHelper.ts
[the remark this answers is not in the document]
maghirardelli: Yup! Gotten this comment like 5 times in this PR now so I think I refactored all of the places where I should have used `Record` in the files in this scope. Please let me know on rereview if I missed one though!

workbench-core/accounts/src/services/projectService.ts
maghirardelli: I am good with that. Thanks for the suggestion!

workbench-core/accounts/src/utilities/projectUtils.test.ts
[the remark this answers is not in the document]
aws-tyler: `projects` is only in the order that its in because of the initial definition. If someone were to change that definition, the sort order would be wrong and the test would fail. By sorting your data as a part of the setup for the test, you avoid depending on the initial definition being in the right order.

bhudnell (approved):

workbench-core/accounts/src/services/projectService.ts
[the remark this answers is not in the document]
maghirardelli: synced with tyler--changing to `badRequest`

workbench-core/accounts/src/utilities/projectUtils.test.ts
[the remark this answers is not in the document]
maghirardelli: We will leave as is but if someone needs to add new entities to the list of projects, they need to verify it doesn't break tests when they do so. I don't really expect someone to add a new project in the set up though.

nguyen102 (dismissed):

aws-tyler (approved):

workbench-core/accounts/src/services/projectService.test.ts
maghirardelli: Tyler and I talked about this and I may come back and use this mocking method in the future.

bhudnell (approved):

aws-tyler (approved):

PR #632, opened 2022-11-17
feat: Update and Delete CostCenter APIs

nguyen102:
Issue #, if available:
GALI-1780

Description of changes:

* Update Cost Center API
* Delete Cost Center API

**Tests**
* Tested Update and Delete Cost Center API with POSTMAN
* Tested that API throws an error if trying to delete a Cost Center that has an API attached

Checklist:


workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -34,6 +36,67 @@ export default class CostCenterService {
...
+   * Update costCenter
+   * @param request - request for updating cost center
+   * @returns CostCenter object with updated attributes
+   */
+  public async updateCostCenter(request: UpdateCostCenterRequest): Promise<CostCenter> {
+    const currentDate = new Date().toISOString();
+
+    const updatedCostCenter = {
+      name: request.name,
+      description: request.description,
```
maghirardelli, line 81: If the user did not pass `description` in, would this evaluate to `undefined` and update the DDB item as `undefined` description?

workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -34,6 +36,67 @@ export default class CostCenterService {
...
+      .get(buildDynamoDBPkSk(request.id, resourceTypeToKey.costCenter))
+      .execute()) as GetItemCommandOutput;
+
+    if (getCostCenterResponse.Item === undefined) {
+      throw Boom.notFound(`Could not find cost center ${request.id}`);
+    }
+
+    try {
+      await this._dynamoDbService
+        .update(buildDynamoDBPkSk(request.id, resourceTypeToKey.costCenter), {
```
maghirardelli, line 62: Your use of update will break when you merge in `develop` since I took out the `unmarshall` from `Updater.execute()`. When you need to fix this, just replace usage of `update(....).execute()` with `updateExecuteAndFormat(...)`. See https://github.com/aws-solutions/solution-spark-on-aws/pull/635 for examples of the refactor.

solutions/swb-app/src/costCenterRoutes.ts
```diff
@@ -23,6 +27,30 @@ export function setUpCostCenterRoutes(router: Router, costCenterService: CostCen
     })
   );
 
+  router.put(
```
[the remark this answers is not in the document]
nguyen102: This is TBD. Currently the other APIs are set as `PUT`. Once Tyler create the one pager about PUT vs PATCH, we'll decide which HTTP method to use. At that point, we'll update all of the APIs at once.

solutions/swb-app/src/costCenterRoutes.ts
```diff
@@ -23,6 +27,30 @@ export function setUpCostCenterRoutes(router: Router, costCenterService: CostCen
...
+    '/costCenters/:id/softDelete',
+    wrapAsync(async (req: Request, res: Response) => {
+      const deleteCostCenterRequest = { id: req.params.id };
+      const validatedRequest = validateAndParse<DeleteCostCenterRequest>(
+        DeleteCostCenterRequestParser,
+        deleteCostCenterRequest
+      );
+      res.status(204).send(await costCenterService.softDeleteCostCenter(validatedRequest));
```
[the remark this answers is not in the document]
nguyen102: As discussed offline, this should be a 204 because the user just needs to know whether the resource was deleted or not. We don't need to return the underlying details of how we implement soft delete. In our case our implementation of soft delete is not the standard approach and will confuse user if we return the object.

workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -34,6 +36,67 @@ export default class CostCenterService {
     this._dynamoDbService = dynamoDbService;
   }
 
+  /**
+   * Soft Delete Cost Center
+   * @param request - request for deleting cost center
+   * @returns void
+   */
+  public async softDeleteCostCenter(request: DeleteCostCenterRequest): Promise<void> {
```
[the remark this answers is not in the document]
nguyen102: Same reasoning from discussing offline

workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -34,6 +36,67 @@ export default class CostCenterService {
...
+    const costCenterHaveProjects = await this._doesCostCenterHaveProjects(request.id);
+    if (costCenterHaveProjects) {
+      throw Boom.conflict(
+        `CostCenter ${request.id} cannot be deleted because it has project(s) associated with it`
+      );
+    }
+
+    const getCostCenterResponse = (await this._dynamoDbService
```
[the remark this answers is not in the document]
nguyen102: Good idea, implementing that.

workbench-core/accounts/src/services/costCenterService.ts
nguyen102: I just double check, and this works as expected. If a user does not pass in a `description` the field does not get changed in DDB.

nguyen102: It's because of this setting
https://github.com/aws-solutions/solution-spark-on-aws/blob/22c4e9677dd0c7c89d47ba98fda748b27422e849/workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts#L319

workbench-core/accounts/src/services/costCenterService.ts
nguyen102: This is good to know. Let me add a TODO here.

workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -34,6 +36,67 @@ export default class CostCenterService {
...
+    const updatedCostCenter = {
+      name: request.name,
+      description: request.description,
+      updatedAt: currentDate
+    };
+
+    let response;
+    try {
+      response = await this._dynamoDbService
+        .update(buildDynamoDBPkSk(request.id, resourceTypeToKey.costCenter), { item: updatedCostCenter })
```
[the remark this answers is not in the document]
nguyen102: Thank you!

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -80,4 +80,21 @@ export default class ProjectService {
...
     return Promise.resolve({ data: projectsResponse.Items as unknown as Project[] });
   }
+
+  /**
+   * Check whether a CostCenter have any projects associated with it
+   * @param costCenterId - id of CostCenter we want to check
+   * @returns Whether a CostCenter have any projects associated with it
+   */
+  public async doesCostCenterHaveProjects(costCenterId: string): Promise<boolean> {
+    const queryParams: QueryParams = {
```
maghirardelli, line 90: You should still add a `limit` parameter so we have bounded queries. This only matters if one is found so we can use `limit: DEFAULT_API` or `limit: 5` even.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -80,4 +80,21 @@ export default class ProjectService {
...
+      index: 'getResourceByDependency',
+      key: { name: 'resourceType', value: 'project' },
+      sortKey: 'dependency',
+      eq: { S: costCenterId }
+    };
+
+    const associatedProjResponse = await this._aws.helpers.ddb.query(queryParams).execute();
```
maghirardelli, line 97: You should use `getPaginatedItems()` instead of `query().execute()`

solutions/swb-app/src/costCenterRoutes.ts
```diff
@@ -35,7 +41,16 @@ export function setUpCostCenterRoutes(router: Router, costCenterService: CostCen
         DeleteCostCenterRequestParser,
         deleteCostCenterRequest
       );
-      res.status(204).send(await costCenterService.softDeleteCostCenter(validatedRequest));
+
+      async function checkDependency(costCenterId: string): Promise<void> {
```
maghirardelli, line 45: What's the point of passing the function to have it call from within the soft delete method? Why not just call this here?

workbench-core/accounts/src/services/projectService.ts
nguyen102: Good idea. I'll make the limit `1` since we just need to know whether a project exist.

workbench-core/accounts/src/services/projectService.ts
nguyen102: updated

maghirardelli (approved):
Looks good!

PR #635, opened 2022-11-18
fix(ddb): replace old update method which returns mutated type

maghirardelli:
Issue #, if available: this is required to allow for proper testing (and merging) of GALI-1738

Description of changes:
I added a new method to `DynamoDBService` that calls the existing `update()` method as well as `execute()` to send the update command and applies proper formatting to the response. I also refactored the old uses of `ddb.update()` to `ddb.updateExecuteAndFormat()`.

Checklist:


workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -271,6 +273,24 @@ export default class DynamoDBService {
     return query;
   }
 
+  /**
+   * Places a DDB Update call and formats the response.
+   *
+   * @param updateParams - {@link UpdateParams} object to pass to the update request
+   * @returns a {@link UpdateUnmarshalledOutput} item where Attributes is unmarshalled
+   */
+  public async updateExecuteAndFormat(updateParams: UpdateParams): Promise<UpdateUnmarshalledOutput> {
```
aws-tyler, line 282: This name is kind of a mouthfull. What about something like 'updateItems'? I'm not sure we want the function names or types to have details about the inner implementation. 

Same thing goes for the return object. Maybe 'UpdateItemsResponse'?

maghirardelli: This is only a single update item call so plural wouldn't work. I wanted to make it easy to differentiate between this function and normal update. Looking at `.update()` and `.updateItem()`, I would not know the difference or which one to use.

Open to other suggestions that still easily paint the difference.

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.test.ts
```diff
@@ -1535,6 +1536,33 @@ describe('DynamoDBService', () => {
       expect(generatedParams).toEqual(expectedParams);
     });
   });
+
+  describe('updateExecuteAndFormat', () => {
+    let unmarshalledAttributes: Record<string, JSONValue>;
```
sevu1, line 1541: Any reason why it's not `Record<string, string>` ?

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -290,60 +310,44 @@ export default class DynamoDBService {
...
-      metrics?: 'NONE' | 'SIZE';
-      capacity?: 'INDEXES' | 'TOTAL' | 'NONE';
-    }
-  ): Updater {
-    let updater = new Updater({ region: this._awsRegion }, this._tableName, marshall(key));
-    if (params) {
-      if (params.disableCreatedAt) {
+  public update(updateParams: UpdateParams): Updater {
+    let updater = new Updater({ region: this._awsRegion }, this._tableName, marshall(updateParams.key));
+    if (updateParams.params) {
```
sevu1, line 315: Could we do early return here? I have strong opinion on if pyramids:
```
if...
  if...
    if...
      if...     
```
vs
```
if !...
  return...
if !...
   return...
etc...
```

sevu1 (approved):

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.test.ts
maghirardelli: Because the value of the attribute could be a number, a list, etc. Tyler wrote this JSONValue type to capture the possibilities.

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
maghirardelli: Doing it this way only requires a single `return` statement. The levels of the `if`s potentially build on each other so the final `return updater` captures all of this.

sevu1: Agree on second level `if`, but `if (!updateParams.params) return updater` should reduce nesting

sevu1, line 315: Also not a strong option here, but we could do spread to reduce chattiness
```
const { params: { disableCreatedAt, set ... } } = updateParams;
if (set)
  updater = updater.set(set);
```

workbench-core/base/src/aws/helpers/dynamoDB/interfaces/updateParams.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface UpdateParams {
+  key: { [key: string]: unknown };
```
bhudnell, line 7: Can these index signatures be refactored to use `Record`?

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
bhudnell: If you wanted to rename:

Memory is a little hazy, are we looking to deprecate the old `.update().execute()` method in favor of this method? If so, it might make sense to mark the old method as deprecated in TSDoc and point the user to use this method instead. Then the naming might not be as confusing.

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -271,6 +273,24 @@ export default class DynamoDBService {
...
+   *
+   * @param updateParams - {@link UpdateParams} object to pass to the update request
+   * @returns a {@link UpdateUnmarshalledOutput} item where Attributes is unmarshalled
+   */
+  public async updateExecuteAndFormat(updateParams: UpdateParams): Promise<UpdateUnmarshalledOutput> {
+    const result = await this.update(updateParams).execute();
+
+    const unmarshalledResult = {
+      Attributes: result.Attributes ? unmarshall(result.Attributes) : undefined,
+      ConsumedCapacity: result.ConsumedCapacity || undefined,
```
bhudnell, line 287: I would consider replacing `||` with `??`. Using the `OR` operator can lead to bugs where if `ConsumedCapacity` is a number and equals 0, it will set the value to `undefined`. The [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) only uses the right hand side if the left hand side is `undefined` or `null`

bhudnell (approved):

maghirardelli:
The non-blocking comments here I am going to revisit when I have downtime. I want to merge this in now.

PR #641, opened 2022-11-21
feat: check if dataset mounted

sevu1:
Issue #, if available: GALI-1849

Description of changes:

Checklist:

* [X] Have you written new tests for your core changes, as applicable?


solutions/swb-app/src/datasetRoutes.ts
```diff
@@ -85,6 +87,25 @@ export function setUpDSRoutes(
...
+  router.delete(
+    '/datasets/:datasetId',
+    wrapAsync(async (req: Request, res: Response) => {
+      const environments: Environment[] = await environmentService.listEnvironmentsForDataSet(
+        req.params.datasetId
+      );
+      if (environments.length) {
+        const existingEnvironments = environments.map((env) => `${env.name}(${env.id})`).join(', ');
+        throw Boom.badRequest(
+          `Cannot delete dataset because it is associated with environments: ${existingEnvironments}`
```
nguyen102, line 100: I'm thinking this message should be `Cannot delete dataset because it has environments associated with it`. 

We shouldn't return the list of existing environments because this list wil eventually be incorrect. What if there's hundreds of environments associated with a dataset? Will we be able to guarantee that we get all environments?

workbench-core/environments/src/services/environmentService.ts
```diff
@@ -537,4 +541,19 @@ export class EnvironmentService {
 
     await this._aws.helpers.ddb.updateExecuteAndFormat({ key, params: { item: data } });
   }
+
+  public async listEnvironmentsForDataSet(dataSetId: string): Promise<Environment[]> {
```
nguyen102, line 545: I think this should live in the [MetadataService](https://github.com/aws-solutions/solution-spark-on-aws/blob/develop/workbench-core/base/src/services/metadataService.ts). This way environments and datasets don't have to be coupled. I believe @SanketD92  has a proposal for how this information can be stored and retrieved.

solutions/swb-app/src/datasetRoutes.ts
```diff
@@ -85,6 +87,25 @@ export function setUpDSRoutes(
...
+        );
+      }
+
+      await dataSetService.removeDataSet(req.params.datasetId);
```
nguyen102, line 104: Thoughts on passing in a `checkDependency` function into `removeDataset`. That way we can ensure that no one will accidentally use `dataSetService.removeDataSet` incorrectly. 

See [here](https://github.com/aws-solutions/solution-spark-on-aws/blob/feat/costCenter/solutions/swb-app/src/costCenterRoutes.ts#L53) for an example.

workbench-core/environments/src/services/environmentService.ts
nguyen102: At a high level you would call `MetadataService.getRelationships('dataset', datasetId, 'environment')`. This will return a list of environments for dataset. You'll make this call in `datasetRoutes.ts`

SanketD92: This is exactly what we need to do, it's just that we need to move this to the metadataService to make it more versatile.

solutions/swb-app/src/datasetRoutes.ts
sevu1: Agree with message change, however I do think we should return at least partial list of environments - otherwise it won't be customer friendly to check every environment whether it has dataset in question

solutions/swb-app/src/datasetRoutes.ts
sevu1: We can do that, however you can still do `.removeDataset(id, () => {});` and avoid checks. The only way to enforce it is to validate all business rules in dataset service

solutions/swb-app/src/datasetRoutes.ts
sevu1: Updated, since service signature was changed

workbench-core/environments/src/services/environmentService.ts
sevu1: Done

solutions/swb-app/src/datasetRoutes.ts
nguyen102: @bhudnell are you ok with us adding a `checkDependency` param to the method signature of `removeDataset`. It forces the implementer to think about whether any dependent resource needs to be deleted before a dataset is removed. We're using a similar approach [here](https://github.com/aws-solutions/solution-spark-on-aws/pull/641#discussion_r1028583222).

solutions/swb-app/src/datasetRoutes.ts
```diff
@@ -85,6 +87,26 @@ export function setUpDSRoutes(
     })
   );
 
+  // Delete dataset
+  router.delete(
+    '/datasets/:datasetId',
+    wrapAsync(async (req: Request, res: Response) => {
+      const { data: environments } = await metadataService.listDependentMetadata<Environment>(
```
nguyen102, line 94: This doesn't actually return a full environment object right? It'll just return an array of `pk, sk, id`.  Typecasting the response to an `environment` is going to be misleading for developers since the IDE will show that the `environment` object has fields that it doesn't actually have.

nguyen102: What if we define a Zod parser and type for `dataset with environment` metadata, and update the `listDependentMetadata` method to accept a Zod parser and return a Zod type? This could look like [this](https://github.com/aws-solutions/solution-spark-on-aws/blob/890744dd99dc857ebd83c56e20c5b7ac9eef3d4b/solutions/swb-app/src/costCenterRoutes.ts#L61). This will allow us to validate that the object type is correct.

nguyen102: [This](https://github.com/aws-solutions/solution-spark-on-aws/blob/890744dd99dc857ebd83c56e20c5b7ac9eef3d4b/workbench-core/accounts/src/models/costCenters/updateCostCenterRequest.ts) is what a Zod parser and type looks like. In our case the Zod type will just have `id`, `pk`, `sk`, and `createdAt`.

solutions/swb-app/src/datasetRoutes.ts
bhudnell: We discussed offline, MAF is fine with the change. Please make a separate PR for it, merging into develop and be sure to include proper integ and unit testing

workbench-core/base/src/services/metadataService.ts
```diff
@@ -3,8 +3,25 @@
...
+  MAX_API_PAGE_SIZE
+} from '../utilities/paginationHelper';
+import { validateAndParse } from '../utilities/validatorHelper';
+
 export class MetadataService {
-  public constructor() {}
+  private readonly _awsService: AwsService;
+
+  public constructor(awsService: AwsService) {
+    this._awsService = awsService;
```
nguyen102, line 22: We can narrow the scope down by just passing in `ddbService` instead of `awsService`

nguyen102 (dismissed):

solutions/swb-app/src/datasetRoutes.ts
```diff
@@ -86,6 +91,33 @@ export function setUpDSRoutes(
...
+          );
+
+        if (environments.length) {
+          throw Boom.badRequest(`Cannot delete dataset because it has environments associated with it.`);
+        }
+      };
+
+      await dataSetService.removeDataSet(
+        req.params.datasetId,
+        async () => await checkDatasetDependencies(req.params.datasetId)
```
bhudnell, line 115: why this rather than just `checkDatasetDependencies`?

solutions/swb-app/src/validatorHelper.ts
```diff
@@ -22,22 +21,4 @@ function processValidatorResult(validatorResult: ValidatorResult): void {
   }
 }
 
-function validateAndParse<T>(parser: ZodTypeAny, data: unknown): T {
```
nguyen102, line 25: Wait is this supposed to be removed? I think other functions use this method.

bhudnell (dismissed):

solutions/swb-app/src/validatorHelper.ts
sevu1: It was moved so downstream packages can use this method: https://github.com/aws-solutions/solution-spark-on-aws/pull/641/files#diff-5619b93169c6639e8622293c9d180c4ed17a51fa2f9e0eda913aa3fcdf9f7fdd

solutions/swb-app/src/schemas/environmentDatasetMetadataParser.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const EnvironmentDatasetMetadataParser = z
```
nguyen102, line 9: Nitpick: I would name this `DatasetEnvironmentMetadataParser` since DS is the pk and environment is the sk

nguyen102 (commented):
Overall looks good. Just one minor comment and once resolved, I can approve.

nguyen102 (approved):

DaveMadsen (approved):

PR #645, opened 2022-11-22
fix: cost center checkDependency

nguyen102:
Issue #, if available:

Description of changes:
fix: checkDepdendency should return Promise<void> not void

Checklist:

### DIFF
--- a/workbench-core/accounts/src/services/costCenterService.ts
+++ b/workbench-core/accounts/src/services/costCenterService.ts
@@ -44,7 +44,7 @@ export default class CostCenterService {
    */
   public async softDeleteCostCenter(
     request: DeleteCostCenterRequest,
-    checkDependency: (costCenterId: string) => void
+    checkDependency: (costCenterId: string) => Promise<void>
   ): Promise<void> {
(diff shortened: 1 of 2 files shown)


workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -44,7 +44,7 @@ export default class CostCenterService {
    */
   public async softDeleteCostCenter(
     request: DeleteCostCenterRequest,
-    checkDependency: (costCenterId: string) => void
+    checkDependency: (costCenterId: string) => Promise<void>
```
nguyen102, line 47: The [function](https://github.com/aws-solutions/solution-spark-on-aws/blob/develop/solutions/swb-app/src/costCenterRoutes.ts#L45) we're passing in is a `Promise` so the parameter type should be a `Promise` as well.

sevu1 (approved):

workbench-core/accounts/src/services/costCenterService.ts
sevu1: Why compiler didn't complain?

PR #652, opened 2022-11-29
chore: update ETC

FernandoAranda:
Issue #, if available:
GALI-1843

Description of changes:
replace schema vallidations with zod,
update ETC APIs with new schema

Checklist:


workbench-core/environments/src/interfaces/createEnvironmentTypeConfigRequest.ts
```diff
@@ -0,0 +1,30 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const CreateEnvironmentTypeConfigRequestParser = z
+  .object({
+    envTypeId: z.string(),
+    params: z
```
sevu1, line 12: Do we want to keep body as `params` or become its own properties?

workbench-core/environments/src/index.ts
```diff
@@ -12,21 +12,31 @@ import {
 import { SortAttribute, isSortAttribute } from './constants/sortAttributes';
 import EnvironmentTypeHandler from './handlers/environmentTypeHandler';
 import StatusHandler from './handlers/statusHandler';
+import {
+  CreateEnvironmentTypeConfigRequest,
+  CreateEnvironmentTypeConfigRequestParser
+} from './interfaces/createEnvironmentTypeConfigRequest';
```
sevu1, line 18: This is not an interface, I'd move it to `models` or better `schemas` / `parsers`

workbench-core/environments/src/interfaces/environmentTypeConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+  createdAt: z.string(),
+  updatedAt: z.string(),
+  estimatedCost: z.string().optional(),
+  params: z.array(
+    z.object({
+      key: z.string(),
+      value: z.string()
+    })
+  )
+});
```
sevu1, line 23: Why not strict here?

workbench-core/environments/src/interfaces/listEnvironmentTypeConfigsRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const ListEnvironmentTypeConfigsRequestParser = z
+  .object({
+    envTypeId: z.string(),
+    pageSize: z
+      .string()
```
sevu1, line 13: Should we do `. number `?

workbench-core/environments/src/interfaces/updateEnvironmentTypeConfigsRequest.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const UpdateEnvironmentTypeConfigRequestParser = z
+  .object({
+    envTypeId: z.string(),
+    envTypeConfigId: z.string(),
+    params: z
```
sevu1, line 13: Same question - should we move params to make them props?

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
```diff
@@ -15,53 +39,57 @@ import {
...
     // provisioningArtifactId: 'pa-dewjn123',
     params: [],
     updatedAt: '2022-06-17T21:25:24.333Z',
-    sk: `${resourceTypeToKey.envType}#${envTypeId}${resourceTypeToKey.envTypeConfig}#${envTypeConfigId}`,
-    owner: 'owner-123',
+    //sk: `${resourceTypeToKey.envType}#${envTypeId}${resourceTypeToKey.envTypeConfig}#${envTypeConfigId}`,
     description: 'Example config 1',
     id: envTypeConfigId,
-    pk: 'ETC',
+    //pk: `${resourceTypeToKey.envType}#${envTypeId}${resourceTypeToKey.envTypeConfig}#${envTypeConfigId}`,
```
sevu1, line 65: Intentional?

workbench-core/environments/src/interfaces/environmentTypeConfig.ts
FernandoAranda: this is used to parse a ddbitem with more props than the ones we need to display, this is not used to parse requests

workbench-core/environments/src/index.ts
FernandoAranda: I've seen other projects use `models` folder, so to follow convention I will change the name of folder to `models`

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: forgot to delete commented line, we don't really need the pk, sk as the object returned does not contain them

workbench-core/environments/src/interfaces/updateEnvironmentTypeConfigsRequest.ts
FernandoAranda: for update I would prefer to separate the update props, that would save some validations, 
if I put all in the same level I would have to add undefined validations to avoid updating a prop that was not sent to the request, for create it makes sense to have it all in the same object

workbench-core/environments/src/interfaces/createEnvironmentTypeConfigRequest.ts
aws-tyler: If the object gets reused (or we expect it to be reused) maybe pull it out? But if its just a one time thing then leave as is?

workbench-core/environments/src/interfaces/listEnvironmentTypeConfigsRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+      .transform((pageSizeString, ctx) => {
+        const pageSize = parseInt(pageSizeString);
```
aws-tyler, line 15: This looks kind of like the code I wrote. Maybe we abstract this transformer into its own thing so people don't have to copy and past it any time we have a pagination info? Then we could have tests for that specific transformer and wouldn't need to re-test that logic for each individual request.

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
```diff
@@ -2,9 +2,33 @@
  *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *  SPDX-License-Identifier: Apache-2.0
  */
-
-const envTypeConfigId = '40b01529-0c7f-4609-a1e2-715068da5f0e';
-jest.mock('uuid', () => ({ v4: () => envTypeConfigId }));
+const envTypeId = '1b0502f3-121f-4d63-b03a-44dc756e4c20';
+const envType = {
```
aws-tyler, line 6: Maybe `const envType: EnvironmentType = ...` so we get all the type checking?

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
```diff
@@ -2,9 +2,33 @@
...
+  description: 'An Amazon SageMaker Jupyter Notebook',
+  id: envTypeId,
+  pk: `ET#${envTypeId}`,
+  productId: 'prod-dwqdqdqdwq2e3',
+  type: 'sagemaker'
+};
+const uuid = '40b01529-0c7f-4609-a1e2-715068da5f0e';
+const envTypeConfigId = `etc-${uuid}`;
+jest.mock('uuid', () => ({ v4: () => uuid }));
+jest.mock('./environmentTypeService', () => {
```
aws-tyler, line 27: I've seen Prototype mocking as well, which feels a little cleaner because everything doesn't need to be hoisted to the top of the file. Doing prototype mocks in the `beforeEach` for that `describe` also means that its scoped instead of global. Do you think it would make sense to replace the module mocking with prototype mocking here?

workbench-core/environments/src/interfaces/environmentTypeConfig.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const EnvironmentTypeConfigParser = z.object({
```
aws-tyler, line 9: I noticed we're not adding `dependency` or `resourceType` here even though they're on the ddb object: should we be?

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
```diff
@@ -15,53 +39,57 @@ import {
...
+  const ddbMock = mockClient(DynamoDBClient);
+
   const envTypeConfig = {
     createdAt: '2022-06-17T16:28:40.360Z',
-    updatedBy: 'owner-123',
-    createdBy: 'owner-123',
     name: 'config 1',
-    allowRoleIds: [],
-    resourceType: 'envTypeConfig',
+    //
```
aws-tyler, line 58: still need this?

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
```diff
@@ -224,14 +244,17 @@ describe('environmentTypeConfigService', () => {
...
-        createParams
-      );
+        params: createParams
+      });
 
       // CHECK
-      expect(actualResponse).toEqual(envTypeConfig);
+      expect(actualResponse).toEqual({
+        ...envTypeConfig,
+        createdAt: actualResponse.createdAt,
```
aws-tyler, line 255: shouldn't `envTypeConfig` already have `createdAt` and `updatedAt` on it?

workbench-core/environments/src/services/environmentTypeConfigService.ts
```diff
@@ -57,31 +40,26 @@ export default class EnvironmentTypeConfigService {
     envTypeId: string,
     envTypeConfigId: string
   ): Promise<EnvironmentTypeConfig> {
-    const response = await this._aws.helpers.ddb
-      .get(this._buildEnvTypeConfigPkSk(envTypeId, envTypeConfigId))
-      .execute();
+    const response = await this._dynamoDbService.get(this._buildEnvTypeConfigPkSk(envTypeConfigId)).execute();
     const item = (response as GetItemCommandOutput).Item;
-    if (item === undefined) {
+    if (item === undefined || (item.dependency as unknown as string) !== envTypeId) {
```
aws-tyler, line 45: why do we need the `as unknown as string` here?

aws-tyler (approved):

workbench-core/environments/src/interfaces/updateEnvironmentTypeConfigsRequest.ts
FernandoAranda: I just found out I dont need the validations, the update method already validates undefined values, I will change it to be own properties

workbench-core/environments/src/interfaces/listEnvironmentTypeConfigsRequest.ts
FernandoAranda: I am following the convention used by other implementations, I think we do this to have more control over the validation message, any thoughts @nguyen102  @aws-tyler

workbench-core/environments/src/interfaces/environmentTypeConfig.ts
FernandoAranda: this is the object we return as response for a get, we dont want to send those fields as part of the response

workbench-core/environments/src/interfaces/listEnvironmentTypeConfigsRequest.ts
aws-tyler: The type should match whatever the type should be in the request. If it is coming in as a string (like how it does via query params) thats where the transformation comes in.

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: I was not returning the ddb response but the object created, meaning that created and updated at would be always different, I changed this in the last commit and fixed this issue

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: removed comments in last commiit

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: added spy on method to receive same result

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: added in last commit

workbench-core/environments/src/interfaces/createEnvironmentTypeConfigRequest.ts
FernandoAranda: moved to top level in last commit

workbench-core/environments/src/index.ts
FernandoAranda: changed folder to models to follow convention

workbench-core/environments/src/interfaces/updateEnvironmentTypeConfigsRequest.ts
FernandoAranda: moved props to top level

workbench-core/environments/src/services/environmentTypeConfigService.test.ts
FernandoAranda: comments deleted

workbench-core/environments/src/services/environmentTypeConfigService.ts
FernandoAranda: dependency is not part of the response , so I didn't really wanted to add another parser for ddbItem, and just calling item.dependency will make compiler complain

aws-tyler (dismissed):

workbench-core/environments/src/interfaces/listEnvironmentTypeConfigsRequest.ts
FernandoAranda: I created a ticket `GALI-2069` for this

workbench-core/environments/src/services/environmentTypeConfigService.ts
FernandoAranda: added getItem method to handle this and avoid using `unknown`

aws-tyler (approved):

solutions/swb-app/src/environmentTypeConfigRoutes.ts
```diff
@@ -61,37 +51,28 @@ export function setUpEnvTypeConfigRoutes(
...
+      const envTypeConfig = await environmentTypeConfigService.listEnvironmentTypeConfigs(
+        listEnvTypeConfigRequest
+      );
+      res.send(envTypeConfig);
     })
   );
 
   // Update envTypeConfig
-  router.put(
+  router.patch(
```
sevu1, line 67: Patch is supposed to do partial update, it seems to be more like put method to me

sevu1 (approved):

solutions/swb-app/src/environmentTypeConfigRoutes.ts
FernandoAranda: its a partial, the update params are optional , so , not sending all props will filter on `update` method and ignore undefined values

workbench-core/environments/src/interfaces/environmentTypeConfig.ts
nguyen102: We should be removing those fields because they're underlying implementation details of our DB. We can't pass those fields to the API users and remove those fields later on because that will be a breaking change. It's best to not pass those fields to the response now. Also we've already started removing those fields from the response of other resource entities. It'll be inconsistent if we make an exception in this case.

bhudnell (approved):

PR #678, opened 2022-12-08
fix(swb): fix accountId needed for using an existing VPC and update AccessLog Cfn Export name

kpark277:
BREAKING CHANGE: UI stack uses previous Cfn Export name, which requires UI stack to be deleted prior to updating swb-reference

Issue #, if available:

Description of changes:
* Update account id to be grabbed from env so that existing VPC can be used in SWB
* Update Cfn Export name for Access Log Bucket to allow for multiple installations of SWB in the same region in the same account

Checklist:

### DIFF
--- a/solutions/swb-reference/src/SWBStack.ts
+++ b/solutions/swb-reference/src/SWBStack.ts
@@ -117,8 +117,11 @@ export class SWBStack extends Stack {
       ALB_INTERNET_FACING
     } = getConstants();
 
+    const MAIN_ACCT_ID: string = process.env.CDK_DEFAULT_ACCOUNT || '';
+
     super(app, STACK_NAME, {
       env: {
+        account: MAIN_ACCT_ID,
         region: AWS_REGION
       }
     });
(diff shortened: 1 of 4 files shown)


solutions/swb-reference/src/SWBStack.ts
```diff
@@ -117,11 +117,22 @@ export class SWBStack extends Stack {
...
     } = getConstants();
 
-    super(app, STACK_NAME, {
-      env: {
-        region: AWS_REGION
-      }
-    });
+    const MAIN_ACCT_ID: string = process.env.CDK_DEFAULT_ACCOUNT || '';
+
+    if (MAIN_ACCT_ID) {
```
aws-tyler, line 122: do we need to do an `if` here? would `account: process.env.CDK_DEFAULT_ACCOUNT || undefined` work?

aws-tyler (approved):

aws-tyler:
Any tests we can add to verify everything is working as expected?

FernandoAranda (approved):

aws-tyler (approved):

solutions/swb-reference/src/SWBStack.ts
kpark277: Good call, fixed.

PR #679, opened 2022-12-09
feat(projects): update project API

maghirardelli:
Issue #, if available: GALI-1923

Description of changes:
Add an Update Project API
also got rid of an old comment in the list parser since i saw it

Integ tests to come

Checklist:


workbench-core/accounts/src/services/projectService.test.ts
```diff
@@ -1494,6 +1496,284 @@ describe('ProjectService', () => {
     });
   });
 
+  describe('updateProject', () => {
+    let projectName: string = project1.name;
```
aws-tyler, line 1500: jest recommends keeping all assignments in the `beforeEach` to prevent any race conditions or unexpected sequencing when prepping the test data. Can you do that for these?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -254,6 +255,55 @@ export default class ProjectService {
...
+   *
+   * @param request - a {@link UpdateProjectRequest} object that contains the id of the project to update
+   *                  as well as the new field values
+   * @returns a {@link Project} object that reflects the changes requested
+   */
+  public async updateProject(request: UpdateProjectRequest): Promise<Project> {
+    const { projectId, updatedValues } = request;
+
+    // disregard empty strings as valid input
+    if (updatedValues.name !== undefined && _.isEmpty(updatedValues.name)) {
```
aws-tyler, line 269: We might want these validation checks in the parser because the parser shouldn't be outputting invalid request object. What do you think?

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -254,6 +255,55 @@ export default class ProjectService {
...
+      updatedValues.name = undefined;
+    }
+    if (updatedValues.description !== undefined && _.isEmpty(updatedValues.description)) {
+      updatedValues.description = undefined;
+    }
+
+    // verify at least one attribute is being updated
+    if (!updatedValues.name && !updatedValues.description) {
+      throw Boom.badRequest('You must supply a new nonempty name and/or description to update the project.');
```
aws-tyler, line 278: I see that we're doing it elsewhere, but the service layer shouldn't be throwing network errors. This is where you'd throw an error from the service and do a try/catch in the controller to handle the response we want to send back to the network.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -254,6 +255,55 @@ export default class ProjectService {
...
+    }
+
+    // if updating name, verify it is not in use and project exists
+    if (updatedValues.name) {
+      await Promise.all([this._isProjectNameInUse(updatedValues.name), this.getProject({ projectId })]);
+    } else {
+      // else, verify project still exists
+      await this.getProject({ projectId });
```
aws-tyler, line 286: I think we can refactor this so that we check first if the project exists, and then if it does, check if the name is being updated is valid. That way we don't need to worry about what might happen in an `else` clause in the future if changes get made.

workbench-core/accounts/src/services/projectService.ts
```diff
@@ -254,6 +255,55 @@ export default class ProjectService {
...
+    }
+
+    // update project DDB item
+    try {
+      const updateResponse = await this._aws.helpers.ddb.updateExecuteAndFormat({
```
aws-tyler, line 291: the dynamo db service should be passed into through the constructor so that it is clear `ProjectService` depends on the dynamo db service. Accessing it through `this._aws.helpers` is an anti-pattern because it hides the relationship and also makes this class vulnerable to any changes in the `AwsService` class and `Helpers` class

aws-tyler (approved):

workbench-core/accounts/src/services/projectService.test.ts
maghirardelli: Yup. I just wanted to give them defaults but I can do that in the `beforeEach` I suppose.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Is this something I should fix now or can I leave it as a room for improvement item since this is consistent with most of the codebase and this service in particular? Happy to sync about refactoring the file as a whole but to my knowledge, this is what we are mostly doing in the services. I do see your point though, but my PR might not be the avenue for the fix.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: The design doc specified doing the checks in parallel. Your way works too but makes the code more sequential at the benefit of not having a condition in the code.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: This is how this service is currently interacting with DB Service. I hear what you are saying, but again, I think this is an opportunity for a room for improvement item for the entire file and not just this method. The refactor can come later if we decide to make our services accept a db service param.

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Going to try using zod [.refine](https://zod.dev/?id=refine) for this

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Going to refactor for all of project service before merging to develop

workbench-core/accounts/src/services/projectService.ts
maghirardelli: Going to refactor for all of project service before merging to develop

aws-tyler (approved):

PR #695, opened 2022-12-14
chore: make Account tests use environment variables

aws-tyler:
Description of changes:
`region` and `table` are not the same in every environment. This uses the environment variables instead of hard-coded values.


sevu1 (approved):

insignias:
@aws-tyler Just to be safe did you deploy and run the integration test ?

insignias:
> > @aws-tyler Just to be safe did you deploy and run the integration test ?
> 
> Only on my personal account, which I believe is what has been causing most of the problems because the two accounts aren't set up the same. Is it ok to use the integration test env to test against from my local machine? I was under the impression I should only test on my account to prevent any issues due to parallel testing on a single account.

since all the PRs are blocked right now, please got ahead and use the integration test env

PR #697, opened 2022-12-14
feat: enable auditing SWB requests

nguyen102:
Issue #, if available:

Description of changes:

Enable logging/auditing every authenticated API requests. A sample audit log looks like this

Lambda API logs
```
{
  timestamp: 1671057400107,
  awsRequestId: '37c8b6ea-f4a6-4668-b68e-83d3fff067ae',
  logEventType: 'audit',
  action: 'POST /costCenters',
  actor: { uid: '1dc5a73f-efac-4abc-85c6-9826838ab510' },
  source: { ip: '<redacted>' },
  body: {
    name: 'CostCenter-3',
    accountId: 'acc-2272f4a8-1791-419a-8e49-b926f3eb46f9',
    description: 'Description for CostCenter-3'
  }
}

ERROR	Failed to get account for cost center creation: Error: Could not find account acc-2272f4a8-1791-419a-8e49-b926f3eb46f9
```

API GW Access Logs
```
{
    "stage": "dev",
    "requestTime": "14/Dec/2022:22:36:39 +0000",
    "requestId": "26117525-794f-419b-9f61-793c19b266a8",
    "integrationRequestId": "37c8b6ea-f4a6-4668-b68e-83d3fff067ae",
    "status": "400",
    "apiId": "5diaqbnwy2",
    "resourcePath": "/{proxy+}",
    "path": "/dev/costCenters/",
    "resourceId": "zu1kdw",
    "httpMethod": "POST",
    "sourceIp": "<redacted>",
    "userAgent": "PostmanRuntime/7.29.2"
}
```

The API lamdba log has most of the request information except for the response HTTP status. That status will show up in the API GW logs. Customers can connect the two requests together by using the `integrationRequestId` and `awsRequestId`

Checklist:


solutions/swb-app/src/audit/swbAuditLogger.ts
```diff
@@ -0,0 +1,14 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuditEntry, Writer } from '@aws/workbench-core-audit';
+
+export default class SwbAuditLogger implements Writer {
+  public async write(metadata: unknown, auditEntry: AuditEntry): Promise<void> {
+    const updatedAuditEntry = { ...auditEntry };
+    delete updatedAuditEntry.statusCode;
+    console.log(updatedAuditEntry);
```
nguyen102, line 12: By default, this will log to CWL. `console.log` will convert the object to a JSON object. The output looks like this
```
2022-12-14T15:37:50.489Z	8247f646-3048-4a9d-9499-203d33a62cd3	INFO	{
  timestamp: 1671032270489,
  awsRequestId: '8247f646-3048-4a9d-9499-203d33a62cd3',
  logEventType: 'audit',
  statusCode: 201,
  action: 'POST /costCenters',
  actor: { uid: '1dc5a73f-efac-4abc-85c6-9826838ab510' },
  source: { ip: '72.83.25.116, 130.176.179.84' },
  body: {
    name: 'CostCenter-3',
    accountId: 'acc-2272f4a8-1791-419a-8e49-b926f3eb46f9',
    description: 'Description for CostCenter-3'
  }
}
```

workbench-core/audit/src/baseExtractor.ts
```diff
@@ -33,10 +33,14 @@ export const BaseExtractor: Extractor = {
       };
 
       // Sets actor to be the AuthenticatedUser's id.
-      const authenticatedUser: AuthenticatedUser = retrieveUser(res);
-      const actor = {
-        uid: authenticatedUser.id
-      };
+      let actor = { uid: 'user not found' };
```
nguyen102, line 36: @htztomic are we ok with using this approach? With this approach we can still log auth requests like `/login` and `/logout`. This allows the auditing logs to be more complete since all requests will still be logged. Without this change, when the `BaseExtractor` tries to `retrieveUser()` it'll error out and the API request won't complete.

workbench-core/audit/src/auditMiddleware.ts
```diff
@@ -49,10 +49,12 @@ export function WithAudit(
     }
     const extractor = auditConfig.extractor ?? BaseExtractor;
     const metadata: Metadata = extractor.getMetadata(req, res);
-    next();
-    metadata.statusCode = res.statusCode;
+    // TODO: Correctly propogate statusCode. StatusCode is not correct when an error is thrown. The error handler is executed after audit log is written, therefore status code is inaccurate
+    // next();
+    // metadata.statusCode = res.statusCode;
```
nguyen102, line 54: @htztomic Are you ok with me removing this field? I've found that if an error is thrown, the HTTP status code in the logs are not accurate. The actual status code is returned [here](https://github.com/aws-solutions/solution-spark-on-aws/blob/523d92babeea50db8219e2a7a5660de9c58dd287/solutions/swb-app/src/errorHandlers.ts#L22), but it's executed after the logs are written. 

I think this is ok, because customers can still get the correct status code by checking API GW access logs. 

Example API GW access log is shown here
```
```
{
    "stage": "dev",
    "requestTime": "14/Dec/2022:22:36:39 +0000",
    "requestId": "26117525-794f-419b-9f61-793c19b266a8",
    "integrationRequestId": "37c8b6ea-f4a6-4668-b68e-83d3fff067ae",
    "status": "400",
    "apiId": "5diaqbnwy2",
    "resourcePath": "/{proxy+}",
    "path": "/dev/costCenters/",
    "resourceId": "zu1kdw",
    "httpMethod": "POST",
    "sourceIp": "<redacted>",
    "userAgent": "PostmanRuntime/7.29.2"
}
```

Comment (DaveMadsen) [reply workbench-core/audit/src/auditMiddleware.ts:54]: Thanks @nguyen102 . will approve and drop a ticket on the backlog to revisit.

File: workbench-core/audit/src/baseExtractor.ts (hunk 2 above)
Comment (DaveMadsen) [reply workbench-core/audit/src/baseExtractor.ts:36]: will approve, but cut a ticket for @htztomic to review when he returns. Looks like it should be good to me.

**Review by DaveMadsen — DISMISSED** [review]

File: workbench-core/audit/src/baseExtractor.ts (hunk 2 above)
Comment (nguyen102) [reply workbench-core/audit/src/baseExtractor.ts:36]: Thanks David. I'll put the ticket here as reference for the future. `MAFoundation-545`

File: solutions/swb-app/config/jest.config.json [hunk 4]
```diff
@@ -1,6 +1,29 @@
...
+    "<rootDir>/src/environmentTypeConfigRoutes.ts",
+    "<rootDir>/src/environmentTypeRoutes.ts",
+    "<rootDir>/src/generateRouter.ts",
+    "<rootDir>/src/index.ts",
+    "<rootDir>/src/projectRoutes.ts",
+    "<rootDir>/src/staticPermissionsConfig.ts",
+    "<rootDir>/src/staticRouteConfig.ts",
+    "<rootDir>/src/userRoutes.ts",
+
+    "<rootDir>/src/validatorHelper.ts",
```
nguyen102, line 24: There was no unit tests in `swb-app` before I added some for audit. Therefore the coverage was close to 0, and I had to ignore these files in test coverage to pass the coverage threshold. I'll create a SIM task to track the files that needs unit test. Unfortunately I do not have enough bandwidth to add unit tests for all of these files in this PR.

solutions/swb-app/src/generateRouter.ts
```diff
@@ -84,6 +88,17 @@ export function generateRouter(apiRouteConfig: ApiRouteConfig): Express {
   app.use(verifyToken(authenticationService, { ignoredRoutes: staticRoutesIgnored, loggingService: logger }));
   app.use(withAuth(authorizationService, { logger: logger }));
 
+  // Auditing
+  const continueOnError = false;
+  const requiredAuditValues = ['actor', 'source'];
+  // Masking fields 'codeVerifier' and 'code' because they're in the body of '/token' API
+  const fieldsToMask = ['user', 'password', 'accessKey', 'code', 'codeVerifier'];
```
sevu1, line 95: Is it possible to move these into configuration? E.g. JSON file. This way it would be easier to modify if we identify that we don't want to log personal information or secrets

solutions/swb-app/src/audit/swbAuditPlugin.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuditEntry, BaseAuditPlugin, Metadata } from '@aws/workbench-core-audit';
+// import * as ServerlessExpress from '@vendia/serverless-express';
```
sevu1, line 7: Forgot to remove?

workbench-core/audit/src/auditMiddleware.ts
sevu1: @nguyen102 , should we configure logger to be executed the last?

workbench-core/audit/src/baseExtractor.ts
```diff
@@ -33,10 +33,14 @@ export const BaseExtractor: Extractor = {
...
+      try {
+        const authenticatedUser: AuthenticatedUser = retrieveUser(res);
```
sevu1, line 38: Should we provide `isAuthenticated` functionality?

solutions/swb-app/src/audit/swbAuditPlugin.ts
nguyen102: good eye. Removing.

solutions/swb-app/config/jest.config.json
aws-tyler: With the type system verifying our input and output types, I'm not sure there are really that many tests to run on the controllers as long as we keep business logic out of them. My making the controllers "thin" and keeping logic in our services, we can consolidate where all the tests need to live. A unit-testing threshold may not be the best way to tell if we've got sufficient code coverage for `swb-app` like it is in reference and core

solutions/swb-app/src/generateRouter.ts
nguyen102: yep, I'll do that

solutions/swb-app/src/audit/swbAuditExtractor.test.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import { Extractor } from '@aws/workbench-core-audit';
+import { AuthenticatedUser } from '@aws/workbench-core-authorization';
+import { Request, Response } from 'express';
+import SwbAuditExtractor from './swbAuditExtractor';
+
+describe('SwbAuditExtractor', () => {
+  let swbAuditExtractor: Extractor;
+  const mockMethod = 'POST';
```
aws-tyler, line 13: Jest recommends setting up all this data in the `beforeEach` to ensure that you can override things properly (especially in cases where there is async work happening). Not a blocker if you don't have time and everything is working right now but just something to keep in mind if you start seeing weird issues.

solutions/swb-app/src/audit/swbAuditLogger.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        actor: { uid: 'sampleID' },
+        source: { ip: 'sampleIP' },
+        body: { name: 'John' }
+      };
+
+      // OPERATE
+      await swbAuditLogger.write({}, auditEntry);
+
+      // CHECK
+      expect(consoleSpy).toBeCalledWith(auditEntry);
```
aws-tyler, line 31: Checking that the method got called means that if the underlying implementation changes and we don't use `global.console` to do logging anymore, this test will break. Is there a way to test inputs and outputs instead?

solutions/swb-app/src/audit/swbAuditPlugin.ts
```diff
@@ -0,0 +1,27 @@
...
+   * @param auditEntry - {@link AuditEntry}
+   */
+  public async prepare(metadata: Metadata, auditEntry: AuditEntry): Promise<void> {
+    const { context } = getCurrentInvoke();
+    auditEntry.awsRequestId = context.awsRequestId;
+    auditEntry.logEventType = 'audit';
+    auditEntry.action = metadata.action;
+    auditEntry.actor = metadata.actor;
+    auditEntry.source = metadata.source;
+    auditEntry.body = metadata.body && typeof metadata.body === 'object' ? metadata.body : undefined;
```
aws-tyler, line 25: maybe throw some `()` around the argument for the ternary so its a little clearer?

workbench-core/audit/src/baseExtractor.ts
nguyen102: Added this request to `MAFoundation-545`

aws-tyler (dismissed):
Some suggestions but no blockers. Nice job!

solutions/swb-app/config/jest.config.json
nguyen102: Yeah for most of the files, I don't think we need unit tests, but for the file listed on line 24 and 25, we could add unit tests for it.

solutions/swb-app/src/audit/swbAuditExtractor.test.ts
nguyen102: Ah that's good to know. I'm going to move the constants into the test block itself since those variables are only being used in that test. Later on, if other tests need those variables, it can be moved to the `beforeEach` block.

solutions/swb-app/src/audit/swbAuditExtractor.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(() => {
+    swbAuditExtractor = new SwbAuditExtractor();
+  });
+  describe('getMetadata', () => {
+    test('with request body', async () => {
+      const res: Response = {
+        locals: {
+          user: mockUser
+        }
+      } as unknown as Response;
```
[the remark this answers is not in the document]
nguyen102: So the issue is a response requires 80 other fields to be mocked, and in our case we only need to mock the `locals` field. That's why the code uses `unknown`. I borrowed this test for the actual `Audit` package itself. I've found a way to not use `unknown` though. I'll follow the example in the SO answer.
https://stackoverflow.com/a/55115072

solutions/swb-app/src/audit/swbAuditLogger.test.ts
nguyen102: Ah yes, so for this implementation we use `console.log` for logging. If we implement another time of logger, we'll want the unit test to break. We'll want the developer to write a new unit test that captures the new logger being used.

solutions/swb-app/src/audit/swbAuditPlugin.ts
nguyen102: haha, I tried, but the formatter that we use isn't happy with that. It automatically removed the parenthesis.

workbench-core/audit/src/auditMiddleware.ts
nguyen102: I spent a couple of hours trying to get the middleware to be executed last, but I wasn't able to get it working. This would need to be investigated further in MAFoundation-545

sevu1 (approved):

workbench-core/audit/src/auditMiddleware.ts
DaveMadsen: I vaguely remember a discussion on firing the audit logging last at the end of the `next()` calls. I think there was an issue with express where execution terminates on `res.send()`. This is definitely something we can revisit in MAFoundation-545 just to ensure I'm remembering correctly and we didn't miss anything.

DaveMadsen (approved):

solutions/swb-app/config/jest.config.json
nguyen102: For reference, here's the SIM ticket GALI-2122

PR #699, opened 2022-12-15
feat: add addPutItems and change addPutRequests to include other options for ddbService

htztomic:
Description of changes:
Add put requests was not given the ability to configure other optional parameters, moved addPutRequests to addPutItems and change addPutRequests to include conditionals

This is a **BREAKING CHANGE**

Checklist:


workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -450,18 +450,36 @@ export default class DynamoDBService {
   }
 
   public transactEdit(params?: {
-    addPutRequest?: Record<string, unknown>[];
+    addPutRequests?: {
+      item: Record<string, unknown>;
```
aws-tyler, line 454: I think we can probably scope these as `Record<string, JsonValue>` instead of `unknown`

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
```diff
@@ -450,18 +450,36 @@ export default class DynamoDBService {
   }
 
   public transactEdit(params?: {
-    addPutRequest?: Record<string, unknown>[];
+    addPutRequests?: {
```
aws-tyler, line 453: Are these all of the options available or is there a full schema somewhere we could reference so we know we've got all available properties exposed?

aws-tyler (approved):

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
htztomic: JSONValue does not encapsulate Sets which is supported by DynamoDB

workbench-core/base/src/aws/helpers/dynamoDB/dynamoDBService.ts
htztomic: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Put.html

htztomic: With one missing attribute, `ReturnValuesOnConditionCheckFailure`,  but can be added when needed

aws-tyler (approved):

PR #709, opened 2022-12-20
feat: [GALI-2124] data set with environment

aws-tyler:
Issue #, if available:
GALI-2124

Description of changes:
Create a `DataSet With Environment` item on environment creation.

### DIFF
--- a/workbench-core/environments/src/services/environmentService.ts
+++ b/workbench-core/environments/src/services/environmentService.ts
@@ -89,48 +89,48 @@ export class EnvironmentService {
    * @param includeMetadata - If true we get all entries where pk = envId, instead of just the entry where pk = envId and sk = envId
    */
   public async getEnvironment(envId: string, includeMetadata: boolean = false): Promise<Environment> {
-    if (includeMetadata) {
-      const data = await this._dynamoDBService
-        .query({ key: { name: 'pk', value: buildDynamoDbKey(envId, resourceTypeToKey.environment) } })
-        .execute();
-      if (data.Count === 0) {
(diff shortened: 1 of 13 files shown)


workbench-core/environments/src/services/environmentService.ts
```diff
@@ -76,13 +77,10 @@ const defaultEnv: Environment = {
...
 export class EnvironmentService {
-  private _aws: AwsService;
-  private _tableName: string;
+  private _dynamoDBService: DynamoDBService;
 
-  public constructor(constants: { TABLE_NAME: string }) {
-    const { TABLE_NAME } = constants;
-    this._tableName = TABLE_NAME;
-    this._aws = new AwsService({ region: process.env.AWS_REGION!, ddbTableName: TABLE_NAME });
+  public constructor(dynamoDBService: DynamoDBService) {
```
SanketD92, line 82: With this change, we'll also need to update the way this service is being referenced in these:
`backendAPI`, `sagemakerNotebookEnvironmentLifecycleService`, `statusHandler`, `environmentLifecycleHelper`.

aws-tyler: 🤦 thank you. rush build didn't fail which is really frustrating. maybe i forgot to run it after that change, not sure.

SanketD92 (approved):
Changes look good. 

But since this change touched on some other lambdas, could you add some details on the regression testing done in the PR description? 
I think a test to provision a sagemaker instance, performing a stop once it's up, and verifying it automatically reaches "Stopped" state should be enough to cover it all.

PR #719, opened 2022-12-22
feat: [GALI-2097] DataSetAuthService

aws-tyler:
Issue #, if available:
GALI-2097

Description of changes:
Added a mock implementation of the `DataSetsAuthorizationPlugin` and integrated it into the DataSetService


solutions/swb-app/src/dataSets/dataSetPermissionParser.ts
```diff
@@ -0,0 +1,16 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const DataSetPermissionParser = z
+  .object({
+    subject: z.string(),
+    accessLevel: z.union([z.literal('read-only'), z.literal('read-write')])
```
FernandoAranda, line 12: does `z.enum` not work for this?

FernandoAranda (approved):

solutions/swb-app/src/dataSets/dataSetPermissionParser.ts
aws-tyler: The accessLevel type is actualy the union of two string literals, not an enum, so I tried to match that exactly without having to add another parser.

FernandoAranda (approved):

solutions/swb-app/src/dataSets/dataSetPermissionParser.ts
FernandoAranda: Interesting so this translates into `'read-only' | 'read-write'`

bhudnell: `z.enum` seems to be the Zod way of creating a union of string literals. https://github.com/colinhacks/zod#zod-enums

aws-tyler: :-( why can't we just have nice enums like everyone else

PR #733, opened 2022-12-30
feat: Implement isUserAssignedToGroup API

bhudnell:
Issue #, if available:

Description of changes:

Checklist:

* [ ] Have you successfully deployed to an AWS account with your changes?
* [ ] Have you written new tests for your core changes, as applicable?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #732, opened 2022-12-30
feat: enable KMS CMK rotation

janeyuaws:
Issue #, if available:
GALI-1695

Description of changes:

- refactor code to use WorkbenchEncryptionKeyWithRotation

- separate `MainAccountEncryptionKey ` into  `S3DatasetsEncryptionKey` and `S3ArtifactEncryptionKey` for the 2 different buckets.
- update hosting account to attach to `S3DatasetsEncryptionKey` key policy

Checklist:

* [x] Have you successfully deployed to an AWS account with your changes?
* [ ] Have you written new tests for your core changes, as applicable?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #734, opened 2022-12-31
feat: add getIdentityPermissionsByIdentity

htztomic:
Issue #, if available:

Description of changes:

Checklist:


workbench-core/authorization/README.md
```diff
@@ -5,7 +5,11 @@
 # Code Coverage
 | Statements                  | Branches                | Functions                 | Lines             |
 | --------------------------- | ----------------------- | ------------------------- | ----------------- |
+<<<<<<< HEAD
+| ![Statements](https://img.shields.io/badge/statements-93.97%25-brightgreen.svg?style=flat) | ![Branches](https://img.shields.io/badge/branches-100%25-brightgreen.svg?style=flat) | ![Functions](https://img.shields.io/badge/functions-63.56%25-red.svg?style=flat) | ![Lines](https://img.shields.io/badge/lines-93.57%25-brightgreen.svg?style=flat) |
+=======
 | ![Statements](https://img.shields.io/badge/statements-93.24%25-brightgreen.svg?style=flat) | ![Branches](https://img.shields.io/badge/branches-100%25-brightgreen.svg?style=flat) | ![Functions](https://img.shields.io/badge/functions-61.78%25-red.svg?style=flat) | ![Lines](https://img.shields.io/badge/lines-92.77%25-brightgreen.svg?style=flat) |
+>>>>>>> 10c6b71473ef8e9e9c5b8387eccbe0d06b8f47c8
```
bhudnell, line 12: missed merge conflict

solutions/swb-reference/src/SWBStack.ts
```diff
@@ -861,7 +861,7 @@ export class SWBStack extends Stack {
       gsis: [
         {
           indexName: 'getIdentityPermissionsByIdentity',
-          partitionKey: { name: 'Identity', type: AttributeType.STRING },
+          partitionKey: { name: 'identity', type: AttributeType.STRING },
```
bhudnell, line 864: why this change?

workbench-core/example/app/src/routes/dynamicAuthorizationRoutes.ts
```diff
@@ -170,4 +170,15 @@ export function setUpDynamicAuthorizationRoutes(router: Router, service: Dynamic
       }
     })
   );
+  router.get(
+    '/authorization/identitypermissions/identity',
```
bhudnell, line 174: In #733 it was suggested to change to `/authorization/permissions/*`

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -357,4 +357,56 @@ describe('DynamicAuthorizationService', () => {
...
+    let sampleGroupType: IdentityType;
+
+    let sampleAction: Action;
+    let sampleEffect: Effect;
+    let sampleSubjectType: string;
+    let sampleSubjectId: string;
+    let sampleConditions: Record<string, JSONValue>;
+    let sampleFields: string[];
+    let sampleDescription: string;
+    let mockIdentityPermission: IdentityPermission;
```
bhudnell, line 372: a lot of these look like duplicates of `createIdentityPermissions`. Can we extract them to a higher level block to reduce duplication?

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationInputs/getIdentityPermissionsByIdentity.ts
```diff
@@ -3,27 +3,37 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { IdentityPermission, IdentityType } from './identityPermission';
+import { z } from 'zod';
+import { IdentityPermission, IdentityTypeParser } from './identityPermission';
 
+// eslint-disable-next-line @rushstack/typedef-var
+export const GetIdentityPermissionsByIdentityRequestParser = z
```
bhudnell, line 10: Do we need request parsers at the service level?

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
```diff
@@ -163,4 +188,49 @@ export class DDBDynamicAuthorizationPermissionsPlugin implements DynamicAuthoriz
   ): string {
     return `${action}|${effect}|${identityType}|${identityId}`;
   }
+  private _transformItemToIdentityPermission(item: Record<string, JSONValue>): IdentityPermission {
+    const { pk, sk, conditions, fields, description } = item;
```
bhudnell, line 192: What happens if these fields dont exist in the `item` object?

workbench-core/example/app/src/routes/dynamicAuthorizationRoutes.ts
DaveMadsen: It was a suggestion. Not sure if there is any reason why permissions would get further overloaded by some other entity.

solutions/swb-reference/src/SWBStack.ts
insignias: This is a part of this [PR](https://github.com/aws-solutions/solution-spark-on-aws/pull/732/files#diff-d52cbf56a77bfc71120b1669d3a69b00dcc221ef47f74d4b7b34cdd0f6d9fc41) and needs to deployed to our test account first as mentioned [here](https://github.com/aws-solutions/solution-spark-on-aws/pull/732#pullrequestreview-1234866718)

bhudnell: The comment there makes it seem like the change didnt originate there?

insignias: The DynamicAuth DynamoDB GSI was created with partitionKey `Identity`, @htztomic pointed out it should have been `identity`. In order to deploy this change we first need to destroy the DDBTable and then recreate it with correct GSI parameters. Hence it would be better to handle this change in a single PR where we update both the SWBStack and the ExampleStack together.

bhudnell: gotcha, makes sense

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
htztomic: these items are transformed from identity permissions. These values are included in the transactWrite when a permission is written

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationInputs/getIdentityPermissionsByIdentity.ts
htztomic: Yes, it is being inferred into the type. This can be utilized to validate input if necessary.

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
```diff
@@ -60,7 +62,30 @@ export class DDBDynamicAuthorizationPermissionsPlugin implements DynamicAuthoriz
...
+    const queryParams = {
+      index: this._getIdentityPermissionsByIdentityIndex,
+      key: {
+        name: this._getIdentityPermissionsByIdentityPartitionKey,
+        value: identity
+      }
+    };
+    let { data, paginationToken } = await this._dynamoDBService.getPaginatedItems(queryParams);
+
+    while (paginationToken) {
```
aws-tyler, line 76: This isn't something I'm super concerned about but want to get on your radar: This is a potential DoS issue, but only at scale. If someone accidentally created enough permissions, they could time out the call by making this loop run for longer than the response timeout. There is possibly also a response-body-size related strangling, but with serverless architecture that may be less of an issue. Anyway, not something I'm worried about for now, but if there is a way to return this data in a paginated way, we could avoid the issue completely.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
```diff
@@ -229,6 +234,23 @@ export class DynamicAuthorizationService {
...
+  /**
+   * Get all identity permissions associated to a specific identity
+   * @param getIdentityPermissionsByIdentityRequest - {@link GetIdentityPermissionsByIdentityRequest}
+   *
+   * @returns - {@link GetIdentityPermissionsByIdentityResponse}
+   */
+  public async getIdentityPermissionsByIdentity(
+    getIdentityPermissionsByIdentityRequest: GetIdentityPermissionsByIdentityRequest
+  ): Promise<GetIdentityPermissionsByIdentityResponse> {
+    getIdentityPermissionsByIdentityRequest = GetIdentityPermissionsByIdentityRequestParser.parse(
```
aws-tyler, line 246: To help delineate when an item has been parsed/validated and when it hasn't, we've been using `validatedRequest`. Not a blocker though.

solutions/swb-reference/src/SWBStack.ts
aws-tyler: Is this going to be a breaking change for anything that has been deployed recently?

aws-tyler (dismissed):

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
bhudnell: Since it looks like these values are required, what are your thoughts on turning this into a more explicit interface?

aws-tyler: Zod could also do parsing/validation check to ensure everything is present and the expected type, which would help get rid of those `as` statements

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -454,6 +458,20 @@ describe('DynamicAuthorizationService', () => {
...
+      }
+    });
+    const request = {
+      identityId: sampleGroupId,
+      identityType: sampleGroupType
+    };
+
+    const { data } = await dynamicAuthzService.getIdentityPermissionsByIdentity(request);
+    expect(data.identityPermissions).toStrictEqual([mockIdentityPermission]);
+  });
```
bhudnell, line 474: I think this test got separated from its parent `describe` in the merge from develop

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -202,45 +239,12 @@ describe('DynamicAuthorizationService', () => {
...
-    let sampleFields: string[];
-    let sampleDescription: string;
-
-    let mockIdentityPermission: IdentityPermission;
-
-    let mockIdentityPermissions: IdentityPermission[];
+    let auditServiceWriteSpy: jest.SpyInstance;
+    beforeAll(() => {
+      auditServiceWriteSpy = jest.spyOn(auditService, 'write');
+    });
```
bhudnell, line 245: This spy has been moved to the outermost `describe` block

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
bhudnell, line 246: Im not sure its necessary to use zod to validate function params at the service level, but not a blocker

aws-tyler: This is probably way too deep of a comment but 🤷 haha. @htztomic and i were actually discussing this the other day. I think TypeScript's lack of runtime validation is a gaping hole in their "we're a typed language!" pitch and so its worth thinking about validation a little differently than you would with Java or something (where you only need logical validation and not typing validation). Validation should probably happen at all module boundaries because someone could be REALLY COOL and use an `as` or something to override the type system's static check and then break things at run time. Zod helps us close that gap a little bit by doing the runtime check via parsing.

If this were just a controller and service that were paired together and never being reused, then ya its probably overkill, but i think validation at the level the module with interface with other code is good to help us make sure our data is always what we expect it to be at run time. Thoughts?

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
```diff
@@ -163,4 +188,49 @@ export class DDBDynamicAuthorizationPermissionsPlugin implements DynamicAuthoriz
...
+      fields: fields as string[],
+      description: description as string
+    };
+  }
+
+  private _decomposeIdentityPermissionsPartitionKey(partitionKey: string): {
+    subjectType: string;
+    subjectId: string;
+  } {
+    const values = partitionKey.split('|');
```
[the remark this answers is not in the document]
htztomic: delimiter shouldn't change but I can abstract it

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.ts
htztomic: Yeah the DoS attack is interesting, these APIs should not be exposed at the handler level directly anyway.

aws-tyler: API exposure isn't the problem, its the amount of data being retrieved. We're ok for small sets of data, but we'd e ok for big sets of data too if we could return just the page of results instead of the entire aggregation of all results.

htztomic: added paginated results

aws-tyler (approved):

workbench-core/authorization/src/dynamicAuthorization/ddbDynamicAuthorizationPermissionsPlugin.test.ts
```diff
@@ -244,11 +234,20 @@ describe('DDB Dynamic Authorization Permissions Plugin tests', () => {
...
         });
 
-      const { data } = await dynamoDBDynamicPermissionsPlugin.getIdentityPermissionsByIdentity({
+      const { data, paginationToken } =
+        await dynamoDBDynamicPermissionsPlugin.getIdentityPermissionsByIdentity({
+          identityId: sampleGroupId,
+          identityType: sampleGroupType
+        });
+      expect(data.identityPermissions).toStrictEqual([mockIdentityPermission]);
+      expect(paginationToken).toStrictEqual(base64PaginationToken);
```
bhudnell, line 243: I'd recommend testing that the entire received response matches the expected response object rather than testing each field separately. This will help ensure we update tests if we add more fields to the response object

htztomic: Each field can be different per response while others can stay the same, testing an entire response can be a plus but I don't think it is necessary here

bhudnell (approved):

PR #739, opened 2023-01-05
feat: verify group status for addUserToGroup

otlci:
Issue #, if available: MAFoundation-577

Description of changes:

Verification of group status when performing addUserToGroup

Checklist:

### DIFF
--- a/workbench-core/authorization/src/dynamicAuthorization/wbcGroupManagementPlugin.test.ts
+++ b/workbench-core/authorization/src/dynamicAuthorization/wbcGroupManagementPlugin.test.ts
@@ -264,7 +264,20 @@ describe('WBCGroupManagemntPlugin', () => {
   });
 
   describe('addUserToGroup', () => {
-    test('returns added equal to true on succesfull call', async () => {
+    beforeEach(() => {
+      ddbMock.on(GetItemCommand).resolves({
+        Item: {
+          id: {
+            S: groupId
+          },
+          status: {
+            S: 'active'
+          }
+        }
+      });
+    });
+
+    test('returns data about new group assignment on succesfull call', async () => {
       const { data } = await wbcGroupManagementPlugin.addUserToGroup({
         groupId,
         userId,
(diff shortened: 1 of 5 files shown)


workbench-core/authorization/src/dynamicAuthorization/wbcGroupManagementPlugin.test.ts
```diff
@@ -264,7 +264,20 @@ describe('WBCGroupManagemntPlugin', () => {
...
+        Item: {
+          id: {
+            S: groupId
+          },
+          status: {
+            S: 'active'
+          }
+        }
+      });
+    });
```
bhudnell, line 278: I would mock `getGroupStatus` here over mocking DDB

workbench-core/authorization/src/dynamicAuthorization/wbcGroupManagementPlugin.ts
```diff
@@ -111,10 +111,17 @@ export class WBCGroupManagementPlugin implements GroupManagementPlugin {
...
-    // ToDo: This requires a check to ensure status of group isn't in pending_delete
     // ToDo: Audit authenticatedUser which actor is performing operation
 
     try {
+      const {
+        data: { status }
+      } = await this.getGroupStatus({ groupId });
+
+      if (status === 'delete_pending') {
+        throw new ForbiddenError(`Cannot assign user to group '${groupId}'. It is pending delete.`);
```
bhudnell, line 122: Would a `GroupNotFoundError` make more sense here since the group is in the process of deletion?

otlci: I used `ForbiddenError` to be consistent with `setGroupStatus` to indicate that operation is not allowed when group is in invalid state. Should there be  another type (e.g. `InvalidGroupStateError`) to indicate this precisely?

bhudnell: The `ForbiddenError` there was because changing the status from `delete_pending` to `active` is not allowed. `GroupNotFoundError` makes sense to me because the group doesnt exist anymore, as it is currently being deleted

bhudnell (approved):

PR #742, opened 2023-01-06
feat(datasets): add AddDataSetAccessPermission to DataSetService

DaveMadsen:
Issue #, if available:

Description of changes:
Add the ability to add an access permission for a DataSet to the DataSetsService

Checklist:

* [ X] Have you successfully deployed to an AWS account with your changes?
* [X ] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/workbench-core/datasets/src/dataSetService.test.ts
+++ b/workbench-core/datasets/src/dataSetService.test.ts
@@ -7,7 +7,7 @@ jest.mock('@aws/workbench-core-audit');
 jest.mock('@aws/workbench-core-authorization');
 jest.mock('@aws/workbench-core-logging');
 jest.mock('./dataSetMetadataPlugin');
-jest.mock('./WbcDataSetsAuthorizationPlugin');
+jest.mock('./wbcDataSetsAuthorizationPlugin');
 
 import { AuditService, BaseAuditPlugin, Writer } from '@aws/workbench-core-audit';
 import {
--- a/workbench-core/datasets/src/dataSetService.test.ts
+++ b/workbench-core/datasets/src/dataSetService.test.ts
@@ -509,9 +509,12 @@ describe('DataSetService', () => {
       });
     });
     it('throws when an invalid dataset Id is given.', async () => {
-      await expect(dataSetService.getDataSet(mockInvalidId)).rejects.toThrow(
-        Boom.notFound(`Could not find DataSet '${mockInvalidId}'.`)
-      );
+      try {
+        await dataSetService.getDataSet(mockInvalidId);
+      } catch (error) {
+        expect(Boom.isBoom(error, 404)).toBe(true);
+        expect(error.message).toBe(`Could not find DataSet '${mockInvalidId}'.`);
+      }
     });
   });
 
@@ -671,9 +674,12 @@ describe('DataSetService', () => {
         ...mockDataSetAddAccessParams,
         dataSetId: mockInvalidId
       };
-      await expect(dataSetService.addDataSetAccessPermissions(invalidAccessParams)).rejects.toThrow(
-        Boom.notFound(`Could not find DataSet '${mockInvalidId}'.`)
-      );
+      try {
+        await dataSetService.addDataSetAccessPermissions(invalidAccessParams);
+      } catch (error) {
+        expect(Boom.isBoom(error, 404)).toBe(true);
+        expect(error.message).toBe(`Could not find DataSet '${mockInvalidId}'.`);
+      }
     });
   });
 });
--- a/workbench-core/datasets/src/dataSetService.test.ts
+++ b/workbench-core/datasets/src/dataSetService.test.ts
@@ -62,9 +62,13 @@ describe('DataSetService', () => {
   const mockDataSetObject = 'datasetObjectId';
   const mockPresignedSinglePartUploadURL = 'Sample-Presigned-Single-Part-Upload-Url';
   const mockUserId = 'sample-user-id';
+  const mockAuthenticatedUser = {
+    id: mockUserId,
+    roles: []
+  };
+
   const mockDataSetAddAccessParams: AddRemoveAccessPermissionRequest = {
-    authenticatedUserId: mockUserId,
-    roles: [],
+    authenticatedUser: mockAuthenticatedUser,
     dataSetId: mockDataSetId,
     permission: {
       accessLevel: 'read-only',
@@ -317,7 +321,8 @@ describe('DataSetService', () => {
           path: mockDataSetPath,
           awsAccountId: mockAwsAccountId,
           region: mockAwsBucketRegion,
-          storageProvider: s3Plugin
+          storageProvider: s3Plugin,
+          authenticatedUser: mockAuthenticatedUser
         })
       ).resolves.toEqual({
         id: mockDataSetId,
@@ -341,7 +346,8 @@ describe('DataSetService', () => {
           path: 'path',
           awsAccountId: 'accountId',
           region: 'bucketRegion',
-          storageProvider: s3Plugin
+          storageProvider: s3Plugin,
+          authenticatedUser: mockAuthenticatedUser
         })
       ).resolves.toEqual({
         id: mockDataSetId,
@@ -359,13 +365,17 @@ describe('DataSetService', () => {
   describe('removeDataset', () => {
     it('returns nothing when the dataset is removed', async () => {
       await expect(
-        dataSetService.removeDataSet(mockDataSetId, () => Promise.resolve())
+        dataSetService.removeDataSet(mockDataSetId, () => Promise.resolve(), mockAuthenticatedUser)
       ).resolves.not.toThrow();
     });
 
     it('throws when an external endpoint exists on the DataSet.', async () => {
       await expect(
-        dataSetService.removeDataSet(mockDataSetWithEndpointId, () => Promise.resolve())
+        dataSetService.removeDataSet(
+          mockDataSetWithEndpointId,
+          () => Promise.resolve(),
+          mockAuthenticatedUser
+        )
       ).rejects.toThrow(
         new DataSetHasEndpointError(
           'External endpoints found on Dataset must be removed before DataSet can be removed.'
@@ -375,18 +385,22 @@ describe('DataSetService', () => {
 
     it('throws when preconditions are not met', async () => {
       await expect(
-        dataSetService.removeDataSet(mockDataSetId, async () => {
-          await Promise.reject(new Error('Preconditions are not met'));
-        })
+        dataSetService.removeDataSet(
+          mockDataSetId,
+          async () => {
+            await Promise.reject(new Error('Preconditions are not met'));
+          },
+          mockAuthenticatedUser
+        )
       ).rejects.toThrow('Preconditions are not met');
     });
   });
 
   describe('getDataSetMountObject', () => {
     it("throws when called with a name that doesn't exists.", async () => {
-      await expect(dataSetService.getDataSetMountObject('name', 'endPointName')).rejects.toThrow(
-        new Error(`'endPointName' not found on DataSet 'name'.`)
-      );
+      await expect(
+        dataSetService.getDataSetMountObject('name', 'endPointName', mockAuthenticatedUser)
+      ).rejects.toThrow(new Error(`'endPointName' not found on DataSet 'name'.`));
     });
 
     it('returns endpoint attributes when called with a name that exists.', async () => {
@@ -415,7 +429,7 @@ describe('DataSetService', () => {
       });
 
       await expect(
-        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId)
+        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId, mockAuthenticatedUser)
       ).resolves.toEqual({
         name: mockDataSetName,
         prefix: mockDataSetPath,
@@ -449,7 +463,7 @@ describe('DataSetService', () => {
       });
 
       await expect(
-        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId)
+        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId, mockAuthenticatedUser)
       ).rejects.toThrow(new Error('Endpoint has missing information'));
     });
 
@@ -477,14 +491,14 @@ describe('DataSetService', () => {
       });
 
       await expect(
-        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId)
+        dataSetService.getDataSetMountObject(mockDataSetId, mockExistingEndpointId, mockAuthenticatedUser)
       ).rejects.toThrow(new Error('Endpoint has missing information'));
     });
   });
 
   describe('listDataSets', () => {
     it('returns an array of known DataSets.', async () => {
-      await expect(dataSetService.listDataSets()).resolves.toEqual([
+      await expect(dataSetService.listDataSets(mockAuthenticatedUser)).resolves.toEqual([
         {
           id: mockDataSetId,
           name: mockDataSetName,
@@ -499,7 +513,7 @@ describe('DataSetService', () => {
 
   describe('getDataSet', () => {
     it('returns a the details of a DataSet.', async () => {
-      await expect(dataSetService.getDataSet(mockDataSetName)).resolves.toEqual({
+      await expect(dataSetService.getDataSet(mockDataSetName, mockAuthenticatedUser)).resolves.toEqual({
         id: mockDataSetId,
         name: mockDataSetName,
         path: mockDataSetPath,
@@ -510,7 +524,7 @@ describe('DataSetService', () => {
     });
     it('throws when an invalid dataset Id is given.', async () => {
       try {
-        await dataSetService.getDataSet(mockInvalidId);
+        await dataSetService.getDataSet(mockInvalidId, mockAuthenticatedUser);
       } catch (error) {
         expect(Boom.isBoom(error, 404)).toBe(true);
         expect(error.message).toBe(`Could not find DataSet '${mockInvalidId}'.`);
@@ -521,7 +535,13 @@ describe('DataSetService', () => {
   describe('addDataSetExternalEndpoint', () => {
     it('returns the mount string for the DataSet mount point', async () => {
       await expect(
-        dataSetService.addDataSetExternalEndpoint(mockDataSetId, mockAccessPointName, s3Plugin, mockRoleArn)
+        dataSetService.addDataSetExternalEndpoint(
+          mockDataSetId,
+          mockAccessPointName,
+          s3Plugin,
+          mockAuthenticatedUser,
+          mockRoleArn
+        )
       ).resolves.toEqual({
         name: mockDataSetName,
         bucket: mockAccessPointAlias,
@@ -538,6 +558,7 @@ describe('DataSetService', () => {
           mockDataSetWithEndpointId,
           mockExistingEndpointName,
           s3Plugin,
+          mockAuthenticatedUser,
           mockRoleArn
         );
         expect.hasAssertions();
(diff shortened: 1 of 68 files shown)


workbench-core/datasets/src/dataSetService.test.ts
```diff
@@ -114,6 +142,8 @@ describe('DataSetService', () => {
             storageName: mockDataSetStorageName,
             externalEndpoints: [mockExistingEndpointName]
           };
+        } else if (id === mockInvalidId) {
```
aws-tyler, line 145: is it possible to move this logic to specific `describe` sections for each test context so its clear which mock results belong to a specific test?

workbench-core/datasets/src/dataSetService.test.ts
```diff
@@ -463,6 +508,11 @@ describe('DataSetService', () => {
         storageName: mockDataSetStorageName
       });
     });
+    it('throws when an invalid dataset Id is given.', async () => {
+      await expect(dataSetService.getDataSet(mockInvalidId)).rejects.toThrow(
+        Boom.notFound(`Could not find DataSet '${mockInvalidId}'.`)
```
aws-tyler, line 513: I think a `rejects.toThrow` check will only compare the message that gets passed in, not the HTTP error. If you change this to another Boom error does it still pass? I think there is a `checkHTTPError()` method that gets used to verify when an HTTP error gets thrown that you may be able to use here.

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -50,16 +52,33 @@ export class DataSetService {
...
+      const provisioned: DataSet = {
+        ...dataSet,
+        storageType: storageProvider.getStorageType()
+      };
+      const response: DataSet = await this._dbProvider.addDataSet(provisioned);
+      metadata.statusCode = 200;
+      await this._audit.write(metadata, response);
+      return response;
+    } catch (error) {
+      metadata.statusCode = 400;
```
aws-tyler, line 78: Why are we hard-coding this to be a 400? Could it ever be another 4xx or 5xx error?

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -50,16 +52,33 @@ export class DataSetService {
...
+      const { storageProvider, ...dataSet } = request;
+
+      await storageProvider.createStorage(dataSet.storageName, dataSet.path);
+
+      const provisioned: DataSet = {
+        ...dataSet,
+        storageType: storageProvider.getStorageType()
+      };
+      const response: DataSet = await this._dbProvider.addDataSet(provisioned);
+      metadata.statusCode = 200;
```
aws-tyler, line 74: Should this be a 201 since we're creating a resource?

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -50,16 +52,33 @@ export class DataSetService {
...
   public async provisionDataSet(request: CreateProvisionDatasetRequest): Promise<DataSet> {
-    const { storageProvider, ...dataSet } = request;
-
-    await storageProvider.createStorage(dataSet.storageName, dataSet.path);
-
-    const provisioned: DataSet = {
-      ...dataSet,
-      storageType: storageProvider.getStorageType()
+    const metadata: Metadata = {
+      actor: { id: '-', roles: [] },
```
aws-tyler, line 56: is `-` a placeholder value?

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -146,7 +165,24 @@ export class DataSetService {
...
    */
   public async getDataSet(dataSetId: string): Promise<DataSet> {
-    return await this._dbProvider.getDataSetMetadata(dataSetId);
+    const metadata: Metadata = {
+      actor: { id: '-', roles: [] },
+      action: this.getDataSet.name,
+      source: {
+        serviceName: DataSetService.name
+      },
+      requestBody: dataSetId
```
aws-tyler, line 174: Is `dataSetId` a requestBody value even if the param is coming from the url and not the body?

workbench-core/example/infrastructure/integration-tests/support/resources/datasets/dataset.ts
```diff
@@ -28,12 +31,48 @@ export default class Dataset extends Resource {
...
   }
 
+  public async addAccess(requestBody: Record<string, unknown>): Promise<PermissionsResponse> {
+    const randomTextGenerator = new RandomTextGenerator(this._settings.get('runId'));
+    const response: AxiosResponse = await this._axiosInstance.post(`${this._api}/permissions`, requestBody);
+    const permissionsCreated: PermissionsResponse = response.data;
+
+    const permissions: Permission[] = permissionsCreated.data.permissions.map((p) => {
+      return {
+        action: p.accessLevel === 'read-only' ? 'READ' : 'UPDATE',
```
aws-tyler, line 57: should this be set via a switch that throws an error if an invalid value gets passed into to the method?  since `requestBody` isn't typed we can't guarantee we're getting valid data.

workbench-core/example/infrastructure/integration-tests/support/resources/datasets/dataset.ts
```diff
@@ -28,12 +31,48 @@ export default class Dataset extends Resource {
...
+        effect: 'ALLOW',
+        subjectType: 'DataSet',
+        subjectId: this.id,
+        identityType: p.identityType === 'GROUP' ? 'GROUP' : 'USER',
```
aws-tyler, line 61: same as above

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  afterAll(async () => {
+    await setup.cleanup();
+  });
+
+  describe('addDataSetAccessPermissions', () => {
+    it('Adds a read-only access permission.', async () => {
+      let response = await adminSession.resources.datasets.create({}, true);
+      const dataSetId: string = response.data.id;
+      response = await adminSession.resources.groups.create({}, true);
```
aws-tyler, line 41: we should keep these immutable for safety

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            {
+              identityType: 'GROUP',
+              identity: groupId,
+              accessLevel: 'read-only'
+            }
+          ]
+        }
+      });
+    });
+    it('Adds a read-only and a read-write access permission when read-write is requested.', async () => {
```
aws-tyler, line 64: This confuses me a little bit: how can you have `read-only` AND `read-write` permissions on the same dataset? They're conflicting permissions, right?

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      let response = await adminSession.resources.datasets.create({}, true);
+      const dataSetId: string = response.data.id;
+      response = await adminSession.resources.groups.create({}, true);
+      const { groupId } = response.data;
+      await expect(
+        (adminSession.resources.datasets.children.get(dataSetId) as Dataset).addAccess({
+          permission: {
+            identityType: 'GROUP',
+            identity: groupId,
+            accessLevel: 'read-only'
```
aws-tyler, line 48: i see a lot of copy and pasted strings: do we have constants for the domains specific strings for AuthZ like `GROUP`, `read-only`, `read-write`, etc?

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
```diff
@@ -0,0 +1,177 @@
...
+              accessLevel: 'read-only'
+            }
+          ]
+        }
+      });
+    });
+    it('Adds a read-only and a read-write access permission when read-write is requested.', async () => {
+      let response = await adminSession.resources.datasets.create({}, true);
+      const dataSetId: string = response.data.id;
+      console.log(dataSetId);
```
aws-tyler, line 67: do you still need this?

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  });
+
+  describe('addDataSetAccessPermissions', () => {
+    it('Adds a read-only access permission.', async () => {
+      let response = await adminSession.resources.datasets.create({}, true);
+      const dataSetId: string = response.data.id;
+      response = await adminSession.resources.groups.create({}, true);
+      const { groupId } = response.data;
+      await expect(
+        (adminSession.resources.datasets.children.get(dataSetId) as Dataset).addAccess({
```
aws-tyler, line 44: why do we need the `as` here?

workbench-core/datasets/src/dataSetService.test.ts
DaveMadsen: fun fact, I had things all separated out long long ago and was asked to push it all up front. I don't think I have time to unwind that today.

workbench-core/datasets/src/dataSetService.test.ts
DaveMadsen: interesting, and fix coming in next push.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: In this case, as we will need those for creating the access permissions.
There is another spot where we aren't and for now, won't be passed the authenticated user however someone made `actor` a required field.

DaveMadsen: oh wait, it's optional! I'll just remove it until I'm ready for it.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: I've asked that on other PRs. The response I've received is along the lines of "well, we don't determine a status code when we throw the error, and we leave it as an exercise to the consumer to choose a status code when they return a result. 200 and 400 are just supposed to represent generic success/failure."

I don't like it either. Working from the "we don't do status codes here" argument, I've removed the status code in the next push.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: see previous comment. Expect this to go away.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: changing to params

workbench-core/example/infrastructure/integration-tests/support/resources/datasets/dataset.ts
DaveMadsen: validation added.

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: you mean the responses? In test code? I'll take a look.

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: `children` if of type `Map<string,Resource>`. `Dataset` (in this case) extends `Resource`.

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: I will look into this. I can use constants if we define them in `base` or `datasets`. I prefer not to use them if they are defined in `authz` as that will encourage making the package coupling stronger. I would eventually like to refactor our code to decouple more.

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: I can change this behavior. 

read-only from an authz perspective is simply `READ`.
read-write is `READ` and `UPDATE` (and if you need them, probably `CREATE`/`DELETE`, but not needed at this juncture for our purposes).

I can create a private method to manage the conversion from READ to read-only and READ/UPDATE to read-write. I originally just used `UPDATE` to capture read-write, but somebody else pushed back on that.

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: nope

aws-tyler (dismissed):

workbench-core/example/infrastructure/integration-tests/tests/datasets/accessPermissions.test.ts
DaveMadsen: okay. finally! I ended up doing some future proofing and separated out converting the dynamic auth output into datasets obejcts into a private method which needed a bunch of unit testing of its own. It's a bit more robust than what I had too.

But now, the READ permission on read-write access will be ignored and UPDATE will supersede with the read-write response.

aws-tyler (approved):

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -50,16 +52,30 @@ export class DataSetService {
...
+      const { storageProvider, ...dataSet } = request;
+
+      await storageProvider.createStorage(dataSet.storageName, dataSet.path);
+
+      const provisioned: DataSet = {
+        ...dataSet,
+        storageType: storageProvider.getStorageType()
+      };
+      const response: DataSet = await this._dbProvider.addDataSet(provisioned);
+      await this._audit.write(metadata, response);
```
htztomic, line 73: metadata needs to be updated with statusCode

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -50,16 +52,30 @@ export class DataSetService {
...
    */
   public async provisionDataSet(request: CreateProvisionDatasetRequest): Promise<DataSet> {
-    const { storageProvider, ...dataSet } = request;
-
-    await storageProvider.createStorage(dataSet.storageName, dataSet.path);
-
-    const provisioned: DataSet = {
-      ...dataSet,
-      storageType: storageProvider.getStorageType()
+    const metadata: Metadata = {
```
htztomic, line 55: will also need to include an actor

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -146,7 +162,22 @@ export class DataSetService {
    * @returns - the DataSet object associated with that DataSet.
    */
   public async getDataSet(dataSetId: string): Promise<DataSet> {
-    return await this._dbProvider.getDataSetMetadata(dataSetId);
+    const metadata: Metadata = {
+      actor: { id: '-', roles: [] },
```
htztomic, line 166: could this be more specific?

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -313,6 +344,41 @@ export class DataSetService {
...
+        serviceName: DataSetService.name
+      },
+      requestBody: params
+    };
+
+    try {
+      // this will throw if the dataset is not found.
+      await this.getDataSet(params.dataSetId);
+      const response: PermissionsResponse = await this._authzPlugin.addAccessPermission(params);
+      await this._audit.write(metadata, response);
```
htztomic, line 374: status code missing here

workbench-core/datasets/src/models/dataSetPermission.ts
```diff
@@ -3,13 +3,17 @@
...
-import { DataSetsAccessLevel } from './dataSetsAccessLevel';
+import { z } from 'zod';
+import { DataSetsAccessLevelParser } from './dataSetsAccessLevel';
 
-export interface DataSetPermission {
+// eslint-disable-next-line @rushstack/typedef-var
+export const DataSetPermissionParser = z.object({
   /** the user or group associated with the access level */
-  identity: string;
+  identity: z.string(),
```
htztomic, line 12: IdentityPermission is defined with this being identityId

workbench-core/datasets/src/models/dataSetPermission.ts
```diff
@@ -3,13 +3,17 @@
...
   /** the type of subject (group or user) */
-  identityType: string;
+  identityType: z.string(),
```
htztomic, line 14:
```suggestion
  identityType: z.union([z.literal('GROUP'), z.literal('USER')]);,
```

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: If it _needs_ to be here, why is it optional on the metadata object?
I don't know what status code will be returned yet. I am not doing a res.send() here.
200/400 are not necessarily accurate and will make the logs confusing.
what do?

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: Not on this PR, but will come in when this method is updated to add an access permission.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: Not without a late game breaking change to a method we did not claim to be updating.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: see previous comment about status codes being optional and confusing.

workbench-core/datasets/src/dataSetService.ts
htztomic: So it is optional as an input, but the auditService will verify if the audit entry is complete before writing it. This is to ensure all audits are following best practices and creating a valid audit trail. By default, auditService will require actor, source, statusCode, and action

workbench-core/datasets/src/models/dataSetPermission.ts
DaveMadsen: Intentionally avoided at this level.

I leave it up to the implementation of the authorization plugin to choose the identity types appropriate for the authz backend and the specific use case. Those are correct for our reference implementation, but defining that union in at this interface level limits future implementations.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: so you're saying that I will need to make a breaking change on every public API to add auditing?

workbench-core/datasets/src/wbcDataSetsAuthorizationPlugin.ts
```diff
@@ -68,21 +68,18 @@ export class WbcDataSetsAuthorizationPlugin implements DataSetsAuthorizationPlug
...
+    const permissionsCount = permissions.length;
 
-    return {
-      data: {
-        dataSetId: createdPermission.data.identityPermissions[0].subjectId,
-        permissions: newPermissions
-      }
-    };
+    if (permissionsCount !== 1) {
+      throw new InvalidPermissionError(
```
htztomic, line 77: when we expect this error, are we expecting users to delete those permissions?

workbench-core/datasets/src/dataSetService.ts
htztomic: not necessarily, actor is dependent on what you determine is sufficient enough to do an audit trial on. If we require the user that is requesting this information to be the actor then yes, we will need that information.

workbench-core/datasets/src/wbcDataSetsAuthorizationPlugin.ts
DaveMadsen: we are not expecting this error.

htztomic (approved):
LGTM

aws-tyler (approved):

PR #757, opened 2023-01-11
feat: implementation of dynamicAuthorization.deleteGroup

otlci:
Issue #, if available: MAFoundation-552

Description of changes:

Checklist:


workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
```diff
@@ -161,9 +163,63 @@ export class DynamicAuthorizationService {
...
+          identityPermissions
+        });
+      } catch (error) {
+        metadata.statusCode = 400;
+        await this._auditService.write(metadata, error);
+        throw new RetryError(error.message);
+      }
+    }
+
+    try {
```
otlci, line 210: @bhudnell : We need to remove the `delete_pending` status once everything is completed gone right?

otlci: I'm not sure I understand, is status stored separately from group? I thought it's group attribute so as group is deleted it gets removed as well. @htztomic ?

workbench-core/example/app/src/routes/dynamicAuthorizationRoutes.ts
```diff
@@ -106,6 +106,28 @@ export function setUpDynamicAuthorizationRoutes(router: Router, service: Dynamic
...
+          authenticatedUser: res.locals.user
+        });
+
+        res.status(200).send(response);
+      } catch (error) {
+        if (isGroupNotFoundError(error)) {
+          throw Boom.notFound(error.message);
+        }
+        if (isRetryError(error)) {
+          throw Boom.serverUnavailable('Failed to delete group permissions. Please request a retry');
```
otlci, line 124: @bhudnell : Can we put a more descriptive error message here?
Updated. I'm not 100% sure about messaging so if there are better suggestions I can update it to something else.

workbench-core/example/infrastructure/integration-tests/tests/dynamicAuthorization/group.test.ts
```diff
@@ -22,13 +22,14 @@ describe('dynamic authorization group integration tests', () => {
...
   beforeEach(() => {
-    expect.hasAssertions();
     fakeUserId = '00000000-0000-0000-0000-000000000000';
     fakeGroupId = '00000000-0000-0000-0000-000000000000';
   });
 
   beforeAll(async () => {
     adminSession = await setup.getDefaultAdminSession();
+
+    expect.hasAssertions();
```
bhudnell, line 32: I believe `expect.hasAssertions` should be in the `beforeEach` block

workbench-core/example/app/src/routes/dynamicAuthorizationRoutes.ts
bhudnell: lgtm

workbench-core/example/infrastructure/integration-tests/tests/dynamicAuthorization/group.test.ts
```diff
@@ -289,4 +290,21 @@ describe('dynamic authorization group integration tests', () => {
...
   });
+
+  describe('deleteGroup', () => {
+    it('deletes existing group', async () => {
+      const {
+        data: { groupId }
+      } = await adminSession.resources.groups.create();
+      const group = adminSession.resources.groups.group(groupId);
+
+      expect(() => group.delete()).not.toThrow();
```
bhudnell, line 301: we can check that the response is what we expect now that we're returning something

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
```diff
@@ -212,9 +214,107 @@ describe('DynamicAuthorizationService', () => {
...
+          source: auditSource,
+          action: auditAction,
+          requestBody: params,
+          statusCode: 400
+        },
+        mockReturnValue
+      );
+    });
+
+    it('throws RetryError exception when deletion of identity permissions fails', async () => {
```
bhudnell, line 286: this test is slightly confusing to me. our `mockReturnValue` is a generic `Error` but the test name says it should throw a `RetryError`

workbench-core/example/infrastructure/integration-tests/tests/dynamicAuthorization/group.test.ts
otlci: `delete` method is derived from `Resource` which has return type of `Promise<void>`. I can't declare same method name with different return type so other option would be to have secondary `delete` functionality but that feels confusing.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
otlci: Per our discussion this will be resolved in subsequent PR; for now I'm going to add comment with ToDo.

workbench-core/example/infrastructure/integration-tests/support/resources/dynamicAuthorization/group.ts
```diff
@@ -29,6 +31,11 @@ export default class Group extends Resource {
     return this._axiosInstance.get(`${this._api}/is-user-assigned/${userId}`);
   }
 
+  public async delete(): Promise<void> {
+    await super.delete();
+    this._clientSession.removeCleanupTask(`group-${this.id}`);
```
bhudnell, line 36: It seems like this should be the default behavior for all delete requests. Can you put this logic in the `Resource` class?

workbench-core/example/infrastructure/integration-tests/tests/dynamicAuthorization/group.test.ts
bhudnell: lets change the default `delete` to return the promise. It wont change any existing tests' behavior, and will allow us to start testing delete response bodies moving forward

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.test.ts
otlci: Per @htztomic any error thrown by `deleteIdentityPermissions` should be wrapped into `RetryError` so the users know that they can retry the operation.

bhudnell (approved):

PR #758, opened 2023-01-11
feat(swb-ref): add SwbAuthZSubject enum and fix roleUtils to expect l…

kpark277:
…owercased project id

Issue #, if available:

Description of changes:
* Adding enum SwbAuthZSubject
* Updating roleUtils to use lowercase project id

Checklist:

### DIFF
--- a/solutions/swb-reference/src/constants.ts
+++ b/solutions/swb-reference/src/constants.ts
@@ -178,4 +178,21 @@ const dataSetPrefix: string = 'DATASET';
 const endPointPrefix: string = 'ENDPOINT';
 const authorizationGroupPrefix: string = 'GROUP';
 
-export { getConstants, getConstantsWithSecrets, dataSetPrefix, endPointPrefix, authorizationGroupPrefix };
+const enum SwbAuthZSubject {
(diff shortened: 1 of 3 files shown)


solutions/swb-reference/src/constants.ts
```diff
@@ -178,4 +178,21 @@ const dataSetPrefix: string = 'DATASET';
 const endPointPrefix: string = 'ENDPOINT';
 const authorizationGroupPrefix: string = 'GROUP';
 
-export { getConstants, getConstantsWithSecrets, dataSetPrefix, endPointPrefix, authorizationGroupPrefix };
+const enum SwbAuthZSubject {
+  SWB_DATASET = 'SWB_DATASET',
```
aws-tyler, line 182: What about `AuthZSubject` so when we use it, the call looks like`AuthZSubject.SWB_DATASET`? Feels weird to have `swb` present twice since we're in `swb-reference`. `AuthZSubject.DATASET` is my preferred way with the underlying value being `SWB_DATASET` but idk if cognitive overhead of having the value not be the enum name is worth it.

aws-tyler (approved):

PR #777, opened 2023-01-18
feat(datasets): add access permissions at dataset creation

DaveMadsen:
Issue #, if available:

Description of changes:

Checklist:

* [ X] Have you successfully deployed to an AWS account with your changes?
* [ X ] Have you written new tests for your core changes, as applicable?


workbench-core/datasets/src/dataSetService.ts
```diff
@@ -295,15 +313,15 @@ export class DataSetService {
 
       if (
         !targetDS.externalEndpoints ||
-        !_.find(targetDS.externalEndpoints, (ep) => ep === targetEndpoint.name)
+        !_.find(targetDS.externalEndpoints, (ep) => ep === targetEndpoint.id)
```
bhudnell, line 316: this should be name, the name is being stored in ddb rather than the id

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -295,15 +313,15 @@ export class DataSetService {
...
       ) {
         return;
       }
 
       await storageProvider.removeExternalEndpoint(targetEndpoint.name, targetDS.awsAccountId!);
 
-      targetDS.externalEndpoints = _.remove(targetDS.externalEndpoints, (endpoint) => {
-        return endpoint === targetEndpoint.name;
+      targetDS.externalEndpoints = _.filter(targetDS.externalEndpoints, (endpoint) => {
+        return endpoint !== targetEndpoint.id;
```
bhudnell, line 324: should be `.name`

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -766,4 +789,31 @@ export class DataSetService {
     );
     return { data: { mountObject } };
   }
+
+  private async _updateNewDataSetPermissions(
+    dataset: DataSet,
+    request: CreateProvisionDatasetRequest
+  ): Promise<DataSetPermission[]> {
+    if (dataset.id) {
+      const permissions = _.isEmpty(request.permissions) ? [] : request.permissions!;
```
bhudnell, line 798:
```suggestion
      const permissions = request.permissions ?? [];
```

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -766,4 +789,31 @@ export class DataSetService {
...
+      if (
+        !_.some(permissions, (p) => p.identity === request.authenticatedUser.id && p.identityType === 'USER')
```
bhudnell, line 800: If the user explicitly provides permissions, should we be adding permission for the authenticated user?

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -766,4 +789,31 @@ export class DataSetService {
     );
     return { data: { mountObject } };
   }
+
+  private async _updateNewDataSetPermissions(
+    dataset: DataSet,
+    request: CreateProvisionDatasetRequest
+  ): Promise<DataSetPermission[]> {
+    if (dataset.id) {
```
bhudnell, line 797: could we early return here?
```
if (!dataset.id) {
  return [];
}
```

bhudnell (changes requested):
revert the `.id` to `.name` comments and lgtm

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: `.id`! see above.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: I thought about that and this is what I came up with. Let me reach out to Kevin and see what he thinks.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen: sure! I was in a group for years that maintained as "one return per method" rule. Can you tell?

workbench-core/datasets/src/dataSetService.ts
bhudnell: It looks like I missed a few locations, and now have some concerns regarding the change. Either where I made the change or elsewhere where `id` is used should have failed a test somewhere.

bhudnell: I think we should complete the change I made in an earlier PR and swap `externalEndpoint.id` for `externalEndpoint.name` everywhere since `name` is also unique and accessible in `addExternalEndpointForUser/Group` where `is` is not.

DaveMadsen: what do you mean by 'accessible'?
>  accessible in addExternalEndpointForUser/Group where id is not.

bhudnell: in `addExternalEndpointForUser/Group` there is a check to see if the endpoint already exists for the dataset. We cant check `dataset.externalEndpoint` for an endpoint id because the endpoint doesnt have one yet. We can check the `externalEndpoint` list for an endpoint name, because that is passed in as a parameter.

DaveMadsen: either way, if you think this is important, please open a ticket. It really is out of scope for this PR. This PR just made it consistent enough to work again. The ticket should also investigate why the unit tests passed.

workbench-core/datasets/src/dataSetService.ts
bhudnell: If Kevin is fine with this approach, I'll approve

bhudnell (approved):

PR #778, opened 2023-01-19
feat(key-pairs): key pair service and skeleton code

maghirardelli:
Issue #, if available: GALI-1758/2071

Description of changes:
Introduces key pair service (currently in swb-ref but will be moved to workbench core on completion) as well as the app plugins and response/request contracts. The methods are unimplemented but throw errors. No APIs are exposed.

I had to test ignore the KeyPairService file since it has no public implemented code. I will remove that from the pattern to ignore when I implement the first public method within that file.

Checklist:


solutions/swb-app/src/keyPairs/createKeyPairRequest.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+import { z } from 'zod';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const CreateKeyPairRequestParser = z
```
FernandoAranda, line 8: do we want to leave it as non strict? Same for all other validation schemas

solutions/swb-app/src/keyPairs/createKeyPairResponse.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+import { KeyPairParser } from './keyPair';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const CreateKeyPairResponseParser = z.object({
```
FernandoAranda, line 10: do we want to leave it as non strict ?

FernandoAranda: same for all other non strict parsers

FernandoAranda (approved):

solutions/swb-app/src/keyPairs/listKeyPairsRequest.test.ts
```diff
@@ -0,0 +1,87 @@
...
+import JSONValue from '@aws/workbench-core-base/lib/types/json';
+import { ListKeyPairsRequestParser } from './listKeyPairsRequest';
+
+describe('ListKeyPairsRequestParser', () => {
+  let requestObject: Record<string, JSONValue>;
+
+  beforeEach(() => {
+    requestObject = { userId: 'valid-uid' };
+  });
+  describe('when pageSize', () => {
```
aws-tyler, line 15: This page size logic seems to be getting spread to a lot of different requests. If you get some time, it would be worth looking into writing a custom Zod transformer that can encapsulate this logic and be shared with all the other requests that use it.

solutions/swb-reference/src/services/keyPairService.ts
```diff
@@ -0,0 +1,111 @@
...
+    throw new Error('Method not implemented.');
+  }
+
+  /**
+   * This method formats a Key Pair object as a DDB item containing key pair data
+   *
+   * @param keyPair - the {@link KeyPair} object to prepare for DDB
+   * @returns an object containing key pair data as well as pertinent DDB attributes based on key pair data
+   */
+  private _mapToDDBItemFromKeyPair(keyPair: KeyPair): Record<string, string> {
```
aws-tyler, line 87: I know we've done this type of transformation previously, but converting to a `Record<string, string>` means we're losing information about what is actually going into dynamo db. What do you think of having a custom class for DynamoDB objects similar to whats happening in `_mapDDBItemToKeyPair`?

aws-tyler (approved):
I think this is a slightly weird way of slicing up the work because there isn't actually any completed/tested functionality, but you've also got a lot of work here already and I don't think its worth stopping this PR to re-slice the work feature-wise instead of component-wise. Next time I'd suggest only making a PR once one full feature is complete. That first PR will have a lot of the same stuff you've got in here but would probably only have one endpoint's worth of infrastructure and then the entire functionality and testing completed for that endpoint.

solutions/swb-app/src/keyPairs/listKeyPairsRequest.test.ts
maghirardelli: We have a backlog item for this I believe!

solutions/swb-reference/src/services/keyPairService.ts
maghirardelli: Open to it! I'll sync with you!

PR #783, opened 2023-01-20
chore: add AppRegistry Construct, integrate AppRegistry with ExampleS…

insignias:
…tack

Issue #, if available:

Description of changes:
1. Add `WorkbenchAppRegistry` Construct: A CDK Construct to create AppRegistry Application, AttributeGroup, share the Application/AttributeGroup across multiple accounts and associate the stacks with the application
2. Add `WorkbenchAssociateStackToExistingAppRegistryApplication` Construct: A CDK Construct to associate a Stack with an existing AppRegistry Application
3. Integrate AppRegistry with ExampleStack

Checklist:


DaveMadsen (approved):

workbench-core/infrastructure/src/workbenchAppRegistry.ts
```diff
@@ -0,0 +1,129 @@
...
+  private _solutionVersion: string;
+  private _appRegistryApplicationName: string;
+  private _applicationType: string;
+  private _attributeGroupName: string;
+  private _accountIds: string[];
+  private readonly _registryApplication: appreg.Application;
+  private readonly _appRegMap: CfnMapping;
+
+  public constructor(scope: Construct, id: string, props: WorkbenchAppRegistryProps) {
+    super(scope, id);
```
FernandoAranda, line 34: do we have cases where the scope will be different from the stack ?

FernandoAranda: if not we can get the id from the scope

insignias: @FernandoAranda For our use case, when we integrate App-Registry it will be associated with atleast one stack so why not pass it as a scope. But we can revisit this if this becomes a problem.

PR #785, opened 2023-01-20
feat: update listDatasets for authZ

bhudnell:
Issue #, if available:

Description of changes:
- updated `listDataSets` API to only return the datasets the user has access to.

Checklist:


workbench-core/datasets/src/dataSetService.ts
```diff
@@ -245,16 +245,32 @@ export class DataSetService {
...
+      const allDatasetsPermissions = await Promise.all(
+        allDatasets.map((dataset) =>
+          this._getAuthenticatedUserDatasetPermissions(authenticatedUser, dataset.id)
+        )
+      );
+
+      allDatasetsPermissions.forEach((permission, index) => {
+        if (permission.length) {
+          //eslint-disable-next-line security/detect-object-injection
+          response.push(allDatasets[index]);
```
DaveMadsen, line 270: Given you are building `allDatasetsPermissions` using potentially several asynchronous calls, are you guaranteed the indexes in `allDatasetsPermissions` will map directly to the same index in `allDataSets`?

workbench-core/example/infrastructure/integration-tests/tests/datasets/list.test.ts
```diff
@@ -24,15 +25,143 @@ describe('datasets list integration test', () => {
...
+        }
+      });
+      await datasets[2].addAccess({
+        permission: {
+          identity: groupId,
+          identityType: 'GROUP',
+          accessLevel: 'read-only'
+        }
+      });
+      await datasets[4].addAccess({
```
DaveMadsen, line 119: You should verify `read-write` access level also works.

workbench-core/example/infrastructure/integration-tests/tests/datasets/list.test.ts
```diff
@@ -24,15 +25,143 @@ describe('datasets list integration test', () => {
...
+      const { data } = await updatedAdminSession.resources.datasets.get();
+      const dataSetIds = data.map((dataset: DataSet) => dataset.id);
+
+      expect(dataSetIds).toStrictEqual(
+        expect.arrayContaining([datasets[0].id, datasets[2].id, datasets[4].id])
+      );
+      expect(dataSetIds).toStrictEqual(expect.not.arrayContaining([datasets[1].id, datasets[3].id]));
+    });
+
+    it('should return the datasets that the user has permissions on from the groups they are in', async () => {
```
DaveMadsen, line 104: Would like to see the language here updated to specify what we SHOULDN'T see here. Emphasize the negative on these tests or explicitly create some strong negative tests.

workbench-core/example/infrastructure/integration-tests/tests/datasets/list.test.ts
```diff
@@ -24,15 +25,143 @@ describe('datasets list integration test', () => {
...
+              identity: userId,
+              identityType: 'USER',
+              accessLevel: 'read-only'
+            }
+          })
+        )
+      );
+    });
+
+    it('should return an empty array when the user has no permissions on any dataset', async () => {
```
DaveMadsen, line 67: We need at least one test that covers each of the unexpected cases where the user has CREATE, UPDATE, DELETE access (IdentityPermissions will need to be created directly through authz), but not READ access. We also need a case here where a user has the 'ALLOW' effect on one READ permission and a 'DENY' on another.

workbench-core/datasets/src/dataSetService.ts
DaveMadsen, line 270: We can only return a dataset here if the permission is for action 'READ' and effect 'ALLOW'. If any permission is a 'DENY'/'READ', it takes precedent over the 'ALLOW' and we do not return it.

This probably needs to be broken out into a common helper that can handle a single dataset so we get it right for external endpoints (which should be updated in a separate ticket) too.

I know we don't directly support having an UPDATE permission without a READ, we don't support CREATE or DELETE permission, and we don't directly support the DENY effect, but because this is authz and can result in potential information disclosure, we need to do it right!

The other option is we throw an exception and return nothing if we hit an unsupported permission, but not sure if that is really the best route here.

DaveMadsen (changes requested):
This is a great start!
I made some comments. My chief concern is that this is authz related, and we need to get it right. In particular, we can only return datasets for which the user has READ access with effect ALLOW and NO permission for DENY-READ.

workbench-core/datasets/src/dataSetService.ts
```diff
@@ -245,16 +245,32 @@ export class DataSetService {
...
     };
+
     try {
-      const response = await this._dbProvider.listDataSets();
+      const response: DataSet[] = [];
+      const allDatasets = await this._dbProvider.listDataSets();
+
+      const allDatasetsPermissions = await Promise.all(
+        allDatasets.map((dataset) =>
+          this._getAuthenticatedUserDatasetPermissions(authenticatedUser, dataset.id)
```
DaveMadsen, line 263: why not just call isAuthorizedOnSubject() for these?

workbench-core/datasets/src/dataSetService.ts
bhudnell: Yes, since `allDataSetsPermissions` is generated by a map function, the permissions there will correlate to the dataset at the same index.

workbench-core/datasets/src/dataSetService.ts
bhudnell: That function isnt available through the `DataSetsAuthorizationPlugin` so I cant use it.

workbench-core/datasets/src/dataSetService.ts
bhudnell: The functions exposed through the `DataSetsAuthorizationPlugin` dont have the concept of `ALLOW` or `DENY`, just `read-only` and `read-write` so I'm not sure how to implement this without a major rework of the plugin

DaveMadsen (approved):

PR #791, opened 2023-01-25
feat: bootstrap it admin

sevu1:
Issue #, if available: GALI-1930

Description of changes: migrate bootstrap script to use dynamic auth Z

Checklist:


workbench-core/environments/src/postDeployment/authorizationSetup.ts
```diff
@@ -57,15 +81,44 @@ export default class AuthorizationSetup {
...
-      identityPermissions: permissions
-    });
+    try {
+      await this._authService.createIdentityPermissions({
+        identityPermissions: permissions,
+        authenticatedUser: user
+      });
+    } catch (err) {
+      if (isIdentityPermissionCreationError(err)) {
+        console.warn(err);
```
aws-tyler, line 116: Works for now but we should have a post-deployment report that will let you know if something went wrong.

aws-tyler (approved):

workbench-core/environments/src/postDeployment/authorizationSetup.ts
sevu1: Unfortunately currently we don't have a good indication on whether everything worked fine or not. The error from DDB transaction is not propagated properly, so normal situation "permissions already exist" cannot be distinguished from "something went wrong". Screenshot for reference:
image

tagging @htztomic / @bhudnell to see if it's possible to provide better error message

PR #809, opened 2023-02-01
fix: delete non existent groups should be idempotent

htztomic:
Description of changes:
Additional refactoring of inputs to models

Checklist:


kpark277 (approved):

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
```diff
@@ -668,4 +650,73 @@ export class DynamicAuthorizationService {
...
+   * @throws - {@link RetryError} if an error is encountered during deletion
+   */
+  private async _deleteAllIdentityPermissionsByIdentity(params: {
+    identityId: string;
+    identityType: IdentityType;
+    authenticatedUser: AuthenticatedUser;
+  }): Promise<void> {
+    const { identityId, identityType, authenticatedUser } = params;
+    try {
+      const identityPermissions = await this._getAllIdentityPermissionsByIdentity({
```
aws-tyler, line 687: I'm not sure if we're at this level of scale yet, but there is the potential to run out of memory if you have enough identity permissions here. You don't have to change it because I don't think we're going to be getting 50k+ entries , but a paginated delete (aka request a page, fire off a delete, repeat) is usually less resource intensive and allows for a place to check if the operation should be canceled mid-execution.

aws-tyler (approved):

workbench-core/authorization/src/dynamicAuthorization/models/addUserToGroup.ts
```diff
@@ -3,7 +3,7 @@
  *  SPDX-License-Identifier: Apache-2.0
```
DaveMadsen, line 3: Please isolate refactor work such as folder restructuring to a single PR. Do not combine with other work.

DaveMadsen (dismissed):
Approved. In the future, please to not mix "fix:" with refactor work.

workbench-core/authorization/src/dynamicAuthorization/dynamicAuthorizationService.ts
```diff
@@ -668,4 +642,73 @@ export class DynamicAuthorizationService {
...
+      await Promise.all(deleteIdentityPermissionsPromises);
+    } catch (error) {
+      throw new RetryError(error.message);
+    }
+  }
+  /**
+   * Check to see if a group has a status
+   * @param params - groupId required to check
+   */
+  private async _hasGroupStatus(params: { groupId: string }): Promise<boolean> {
```
bhudnell, line 700: Do we need this function now that we're throwing `GroupNotFoundError` in the `getGroupStatus` function?

htztomic: calling getGroupStatus without using the response and waiting to see if it throws feels rather not intuitive. I think keeping this will be clearer and cleaner

bhudnell (approved):

aws-tyler (approved):

PR #829, opened 2023-02-07
test: added refreshAccessToken integration tests

bhudnell:
Issue #, if available:

Description of changes:
- Added integration tests for the authentication package's `refreshAccessToken` API

Checklist:

### DIFF
--- a/workbench-core/example/infrastructure/integration-tests/tests/authentication/routeHandler.test.ts
+++ b/workbench-core/example/infrastructure/integration-tests/tests/authentication/routeHandler.test.ts
@@ -79,6 +79,7 @@ describe('authentication route handler integration tests', () => {
 
       expect(data.loggedIn).toBe(false);
     });
+
     it('should return false for an invalid access token', async () => {
       const { data } = await anonymousSession.resources.authentication.loggedIn({
         access: false,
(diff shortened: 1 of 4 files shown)


workbench-core/example/infrastructure/integration-tests/tests/authentication/routeHandler.test.ts
```diff
@@ -104,4 +107,29 @@ describe('authentication route handler integration tests', () => {
       expect(data.loggedIn).toBe(false);
     });
   });
+
+  describe('refreshAccessToken', () => {
+    it('should return the id token if the access token is successfully refreshed', async () => {
```
DaveMadsen, line 112: Do we need a test for an expired access token?

DaveMadsen (approved):
one comment, not blocking this code. Can be addressed here or another ticket if it needs to be addressed.

workbench-core/example/infrastructure/integration-tests/tests/authentication/routeHandler.test.ts
bhudnell: In theory we probably should. In practice, the refresh token is stored in a cookie that expires and deletes at the same time the token itself expires so this is a very edge case. If we want to test it, it will take a decent amount of effort to figure out how to generate a valid but expired refresh token to test with. So I would say handle it in another ticket.

PR #835, opened 2023-02-09
topic: multiple client ids

otlci:
Issue #, if available:

Description of changes:

Checklist:


workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.test.ts
```diff
@@ -86,7 +88,10 @@ describe('CognitoAuthenticationPlugin tests', () => {
 
   describe('constructor tests', () => {
     it('should throw PluginConfigurationError when the user pool id is invalid. Must match "<region>_<some string>" format', () => {
-      const badUserPoolIdConfig = { ...cognitoPluginOptions, userPoolId: 'badId' };
+      const badUserPoolIdConfig = {
+        ...cognitoPluginOptions,
+        webUiAppClient: { ...cognitoPluginOptions.webUiAppClient, userPoolId: 'badId' }
```
DaveMadsen, line 93: it looks like tests are going through the UI client. This is the client which should have `ALLOW_ADMIN_USER_PASSWORD_AUTH` disabled and should not function with integration tests.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -9,7 +9,7 @@ import {
   DescribeUserPoolClientCommandInput
 } from '@aws-sdk/client-cognito-identity-provider';
 import { CognitoJwtVerifier } from 'aws-jwt-verify';
-import { CognitoJwtVerifierSingleUserPool } from 'aws-jwt-verify/cognito-verifier';
+import { CognitoJwtVerifierMultiUserPool } from 'aws-jwt-verify/cognito-verifier';
```
DaveMadsen, line 12: I think we need multi-user-client, not multi-user-pool.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.test.ts
otlci: Updated.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,45 +29,59 @@ interface TokensExpiration {
...
 
-export interface CognitoAuthenticationPluginOptions {
+interface CognitoAppClient {
   /**
-   * The Cognito domain. Follows the format: "https://\<domain prefix\>.auth.\<region\>.amazoncognito.com"
+   * The Cognito app client IDs.
    */
-  cognitoDomain: string;
-
+  clientId: string | string[];
```
bhudnell, line 36: Each AppClient should only have one client ID associated with it

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,45 +29,59 @@ interface TokensExpiration {
...
   /**
    * The Cognito user pool ID. Follows the format: "\<region\>_\<some string\>"
    */
   userPoolId: string;
```
bhudnell, line 40: Every AppClient should belong to the same user pool

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,45 +29,59 @@ interface TokensExpiration {
...
+   */
+  clientSecret: string;
+}
+
+interface WebUiAppClient extends Omit<CognitoAppClient, 'clientId'> {
+  /**
+   * The Cognito app client IDs.
    */
   clientId: string;
+}
```
bhudnell, line 53: unnecessary with the above clientId change

workbench-core/example/infrastructure/integration-tests/support/setup.ts
```diff
@@ -29,9 +29,7 @@ export default class Setup {
     return session;
   }
 
-  public async createAdminSession(): Promise<ClientSession> {
-    const userPoolId = this._settings.get('ExampleCognitoUserPoolId');
-    const clientId = this._settings.get('ExampleCognitoUserPoolClientId');
+  public async createClientSession(userPoolId: string, clientId: string): Promise<ClientSession> {
     const rootUserNameParamStorePath = this._settings.get('rootUserNameParamStorePath');
     const rootPasswordParamStorePath = this._settings.get('rootPasswordParamStorePath');
```
bhudnell, line 34: This info is from an admin user, so renaming it from `createAdminSession` to a generic `createClientSession` seems misleading to me.

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -68,15 +69,20 @@ const userPoolClientDefaults: UserPoolClientOptions = {
...
 export interface WorkbenchCognitoProps {
   domainPrefix: string;
   websiteUrls: string[];
   userPoolName?: string;
-  userPoolClientName?: string;
   oidcIdentityProviders?: WorkbenchUserPoolOidcIdentityProvider[];
-  accessTokenValidity?: Duration;
-  idTokenValidity?: Duration;
-  refreshTokenValidity?: Duration;
+  userPoolClients?: WorkbenchCognitoUserPoolClientProps[];
```
bhudnell, line 85: The web ui client should have a separate field since it is handled differently than the programmatic access clients

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,45 +29,59 @@ interface TokensExpiration {
...
   /**
    * The Cognito user pool ID. Follows the format: "\<region\>_\<some string\>"
    */
   userPoolId: string;
 
   /**
-   * The Cognito app client ID.
+   * The Cognito app client secret.
+   */
+  clientSecret: string;
```
bhudnell, line 45: Programmatic access clients dont necessarily need client secrets. We just need the client id to add to the JWT verifier

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
otlci: Ticket describes two scenarios, one is passing multiple AppClients from same user pool but second part describes passing various user pool AppClients:

```
Ask for from 
[@thingut](https://amzn-aws.slack.com/team/W017VKKSHS5)
: We need to provide a way for access tokens generated from other user pool clients to pass the verifyToken method.
```

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,45 +29,59 @@ interface TokensExpiration {
...
+  cognitoDomain: string;
+
+  /**
+   * Web UI Cognito App Client information.
+   */
+  webUiAppClient?: WebUiAppClient;
+
+  /**
+   * Cognito App Clients information for token validation supporting ALLOW_ADMIN_USER_PASSWORD_AUTH authentication mode.
+   */
```
bhudnell, line 68: They dont necessarily need to support `ALLOW_ADMIN_USER_PASSWORD_AUTH`, these are just the clients that are also allowed to verify their access tokens

bhudnell (changes requested):
See my comments about refactoring to only a single user pool and splitting the web ui client from the other clients.

workbench-core/example/infrastructure/integration-tests/support/setup.ts
otlci: Renamed to `createRootUserSession` to help clarify the intention

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
otlci: Updated comment and property name.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
bhudnell: The wording of the ask is a little confusing, but they are asking for multiple app clients from the same user pool.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
bhudnell: They are referring to allowing for multiple app clients from the same user pool. The current implementation only allows for a single app client (the web UI app client)

workbench-core/infrastructure/src/workbenchCognito.ts
otlci: I'm not sure I understand since this is for CloudFormation template - webUi client is created with 

```
authFlows: {adminUserPassword: false}
```

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
bhudnell: Looking through the implementation, I think it would be better to remove it entirely. The client secret is only used for the web UI, the other AppClients wont use it at all

otlci: It seems to be used for  webUi client in `_getEncodedClientId` method and by extension by `revokeToken`.

workbench-core/infrastructure/src/workbenchCognito.ts
bhudnell: Since the web UI client is used in a different way than the other clients it should be split out so a user doesnt have to remember where the web UI clientId was in the array of clientIds or if they didnt include it at all

bhudnell: I would also recommend removing `adminUserPassword: true` from the `userPoolClientDefaults` object and force the user to allow that flow explicitly rather than allow it by default.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
bhudnell: The decision to allow users from other user pools authenticate with this service requires further discussion and is outside the scope of this ticket.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
otlci: OK. This means that we won't need `CLIENT_SECRET` variable in `ExampleStack`, correct?

bhudnell: The only client secret needed in the example stack is the secret for the web UI app client

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -51,53 +57,70 @@
...
-      throw new PluginConfigurationError('Invalid Cognito user pool id');
+    const combinedAppClients = [...(webUiAppClient ? [webUiAppClient] : []), ...appClients];
+
+    if (combinedAppClients.length === 0) {
+      throw new PluginConfigurationError('At least one appClient must be provided');
+    }
+
+    const regionMatch = combinedAppClients.find(
+      // eslint-disable-next-line security/detect-unsafe-regex
+      ({ userPoolId }) => !userPoolId.match(/^(?<region>(\w+-)?\w+-\w+-\d)+_\w+$/)
```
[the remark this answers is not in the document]
bhudnell: Thinking about this regex, we can probably remove the `+` and explicitly state how many characters to expect based on the current AWS region codes. The only risk here is we would need to update the regex if there is a new region code that doesnt match our expectations

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -386,11 +422,19 @@
...
   private async _getTokensExpirationinMS(): Promise<TokensExpiration> {
-    const client = new CognitoIdentityProviderClient({ region: this._region });
+    if (!this._webUiAppClient) {
+      throw new PluginConfigurationError('WebUI client not configured');
+    }
+
+    const { userPoolId, clientId } = this._webUiAppClient;
+
+    // eslint-disable-next-line security/detect-unsafe-regex
+    const regionMatch = userPoolId.match(/^(?<region>(\w+-)?\w+-\w+-\d)+_\w+$/);
```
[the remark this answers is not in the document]
bhudnell: Probably worthwhile to move this regex to a constant since its being used multiple places.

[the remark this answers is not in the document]
bhudnell: Same comment as above.

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
otlci: Removed.

PR #839, opened 2023-02-10
feat: disassociate data set from project

aws-tyler:
Adds logic to remove the association between a dataset and project


solutions/swb-app/src/dataSetRoutes.ts
```diff
@@ -119,4 +120,24 @@ export function setUpDSRoutes(router: Router, dataSetService: DataSetPlugin): vo
...
+      });
+
+      await dataSetService.removeAccessForProject(validatedRequest);
+
+      res.status(204).send();
+    })
+  );
 }
+
+//update dataset
```
aws-tyler, line 140: removing

solutions/swb-reference/config/jest.config.json
```diff
@@ -17,15 +17,17 @@
...
     "<rootDir>/src/services/databaseService.ts",
     "<rootDir>/src/services/projectEnvTypeConfigService.ts"
   ],
   "coverageThreshold": {
     "global": {
-      "branches": 70,
-      "functions": 70,
-      "lines": 70,
-      "statements": 70
+      "branches": 50,
```
aws-tyler, line 27: I lowered these because we're getting penalized on the DataSetService methods that are purely pass-throughs to the workbench-core DataSetService. I don't think code coverage is a sensible metric for what we need in `swb-reference`, but i'm happy to discuss if there is opposition to this change.

solutions/swb-reference/integration-tests/support/setup.ts
```diff
@@ -58,6 +58,33 @@ export default class Setup {
     return this._defaultAdminSession;
   }
 
+  public async getSessionForUser(userName: string, password: string): Promise<ClientSession> {
```
aws-tyler, line 61: This is the infrastructure needed to get sessions for users of different roles.

solutions/swb-reference/src/services/dataSetService.ts
```diff
@@ -176,6 +176,63 @@ export class DataSetService implements DataSetPlugin {
     return response;
   }
 
+  public async removeAccessForProject(request: ProjectAccessRequest): Promise<PermissionsResponse> {
+    const projectAdmin = getProjectAdminRole(request.projectId);
+
+    //Make sure you're not removing the access for your project
+    if (request.authenticatedUser.roles.includes(projectAdmin)) {
+      //TODO: Throw a better error
```
aws-tyler, line 184: Missed this. Will update.

solutions/swb-reference/src/services/dataSetService.ts
```diff
@@ -176,6 +176,63 @@ export class DataSetService implements DataSetPlugin {
...
+      type: SwbAuthZSubject.SWB_DATASET,
+      id: request.dataSetId
+    };
+
+    const project: Associable = {
+      type: SwbAuthZSubject.SWB_PROJECT,
+      id: request.projectId
+    };
+
+    console.log(`removeAssociations`);
```
aws-tyler, line 230: remove

solutions/swb-reference/src/services/dataSetService.ts
```diff
@@ -176,6 +176,63 @@ export class DataSetService implements DataSetPlugin {
...
+
+    //Make sure you're not removing the access for your project
+    if (request.authenticatedUser.roles.includes(projectAdmin)) {
+      //TODO: Throw a better error
+      throw new Error("can't remove access from a project admin on a dataset they own");
+    }
+
+    const projectResearcher = getResearcherRole(request.projectId);
+
+    console.log(`removeAuthZPermissionsForDataset`);
```
aws-tyler, line 190: remove

solutions/swb-app/src/dataSets/projectAccessRequestParser.ts
```diff
@@ -12,7 +12,7 @@ export const ProjectAccessRequestParser = z
     authenticatedUser: AuthenticatedUserParser,
     dataSetId: z.string(),
     projectId: z.string(),
-    accessLevel: z.enum(['read-write', 'read-only'])
+    accessLevel: z.enum(['read-write', 'read-only']).optional()
```
maghirardelli, line 15: Is there a default? What happens when nothing is passed?

solutions/swb-reference/config/jest.config.json
maghirardelli: Can we add that file to the ignoreFiles instead?

solutions/swb-reference/integration-tests/support/utils/settings.ts
```diff
@@ -15,6 +15,11 @@ interface Setting {
   terminatedEnvId: string;
   rootUserNameParamStorePath: string;
   rootPasswordParamStorePath: string;
+  researcherUserNameParamStorePath: string;
+  researcherPasswordParamStorePath: string;
+  projectAdminUserNameParamStorePath: string;
+  projectAdminPasswordParamStorePath: string;
+  projectForRolesTesting: string;
```
maghirardelli, line 22: 1) we already have projectId above--could that be repurposed for this as well?
2) do we want our researcher and our project admin to be in different projects so that we could test the boundary better?

solutions/swb-reference/src/services/dataSetService.ts
```diff
@@ -147,7 +147,7 @@ export class DataSetService implements DataSetPlugin {
       permission: {
         identity: projectAdmin,
         identityType: 'GROUP',
-        accessLevel: request.accessLevel
+        accessLevel: request.accessLevel!
```
maghirardelli, line 150: What happened when this value does not exist?

solutions/swb-reference/src/services/dataSetService.ts
```diff
@@ -176,6 +176,60 @@ export class DataSetService implements DataSetPlugin {
...
+    }
+
+    const projectResearcher = getResearcherRole(request.projectId);
+
+    await this._removeAuthZPermissionsForDataset(
+      request.authenticatedUser,
+      SwbAuthZSubject.SWB_DATASET,
+      request.dataSetId,
+      [projectAdmin, projectResearcher],
+      ['READ']
```
maghirardelli, line 194: Why only read? Why not remove write as well?

solutions/swb-app/src/dataSets/projectAccessRequestParser.ts
aws-tyler: There is no default. The value is only required when you're adding a project, not when removing one, and we check to make sure we have the value when we need it.

solutions/swb-reference/integration-tests/support/utils/settings.ts
aws-tyler: Unless things go very poorly, this is all temporary infrastructure that will go away once we can generate accounts at test-time. If things do go poorly and we can't generate at test time, you're right that we'll need two sets of PAs and Researchers that are members of different projects.

solutions/swb-reference/config/jest.config.json
aws-tyler: Unfortunately i don't think we can because there _is_ custom logic in some of the methods in the DataSetService (like adding authz permissions, etc).

maghirardelli (approved):

PR #850, opened 2023-02-15
chore: fix appRegistry Construct

insignias:
Issue #, if available:

Description of changes:

Checklist:


workbench-core/infrastructure/src/workbenchAppRegistry.test.ts
```diff
@@ -59,17 +59,17 @@ describe('WorkbenchAppRegistry Test', () => {
 
     template.hasResourceProperties('AWS::ServiceCatalogAppRegistry::AttributeGroupAssociation', {
       Application: {
-        'Fn::GetAtt': ['AppRegistry968496A3', 'Id']
+        'Fn::GetAtt': ['TestStackApplicationF250E570', 'Id']
```
DaveMadsen, line 62: `F250E570` is really a constant which can be used reliably as a literal here? Same applies to similar on line 65.

DaveMadsen (dismissed):
I saw some odd literals in the test code and left a comment, but it is test code, and they are just different odd literals so I will approve.

workbench-core/infrastructure/src/workbenchAppRegistry.test.ts
insignias: Yes @DaveMadsen we can use these literals here, these don't change unless the code is updated

FernandoAranda (approved):

bhudnell (approved):

PR #852, opened 2023-02-16
fix: updateEnvironment params order was backwards

kpark277:
Issue #, if available:

Description of changes:
* Updated order of params for projectEnvironmentService/Plugin so that Project always comes first.

Checklist:


aws-tyler (approved):
How did this get through tests in the first place? Are there any tests that need to be fixed to catch this sort of change?

PR #858, opened 2023-02-17
fix(swb-reference): fix signed URL fails to download cfn template

DaveMadsen:
Issue #, if available:

Description of changes:
This fixes the reported issue where attempting to utilize a signed URL fails to decrypt artifact bucket contents due to lack of access to the KMS key.

This also adds a little fix to upload the BYON and TGW cfn templates to the artifacts bucket.

Checklist:


solutions/swb-reference/src/SWBStack.ts
```diff
@@ -1090,6 +1092,17 @@ export class SWBStack extends Stack {
             resources: [`arn:aws:kms:${AWS_REGION}:${this.account}:key/*`],
             sid: 'KMSAccess'
           }),
+          new PolicyStatement({
```
maghirardelli, line 1095: Why not combine this with the above policy? And should the above policy be scoped to the same key as this? I believe so but i am not the DS expert.

DaveMadsen: Looks like the above statement is specific to DataSets, but without BYOB enabled, I think there is just the one datasets bucket encrypted with the main account key so it *should* work. The question becomes whether to leave the above statement in for future BYOB or collapse them into one for this release.

maghirardelli: Ah, I see. Does DS not need decrypt and encrypt powers too?

Personally, I do not have a strong preference to combine or leave room for extension. Potentially, combining is safer because it is properly scoped for _current_ feature capability. When we do BYOB, we could always add it in then. Feel free to 16th minute it if you want more feedback from the team though.

maghirardelli (approved):

PR #873, opened 2023-02-20
feat: list data set access permissions

aws-tyler:
Issue #, if available:

Description of changes:
Added endpoint and tests to allow a user to get the permissions associated with a data set so they can see who has access to their dataset

Checklist:

### DIFF
--- a/solutions/swb-app/src/dataSetRoutes.ts
+++ b/solutions/swb-app/src/dataSetRoutes.ts
@@ -15,6 +15,10 @@ import {
 } from './dataSets/createExternalEndpointRequestParser';
 import { DataSetPlugin } from './dataSets/dataSetPlugin';
 import {
+  ListDataSetAccessPermissionsRequest,
+  ListDataSetAccessPermissionsRequestParser
+} from './dataSets/listDataSetAccessPermissionsRequestParser';
+import {
   ProjectAddAccessRequest,
   ProjectAddAccessRequestParser
 } from './dataSets/projectAddAccessRequestParser';
@@ -146,6 +150,21 @@ export function setUpDSRoutes(router: Router, dataSetService: DataSetPlugin): vo
       await dataSetService.removeAccessForProject(validatedRequest);
 
       res.status(204).send();
+    })
+  );
+
+  router.get(
+    '/datasets/:datasetId/permissions',
+    wrapAsync(async (req: Request, res: Response) => {
+      const validatedRequest = validateAndParse<ListDataSetAccessPermissionsRequest>(
+        ListDataSetAccessPermissionsRequestParser,
+        {
+          ...req.query,
+          authenticatedUser: res.locals.user,
+          dataSetId: req.params.datasetId
+        }
+      );
+      res.send(await dataSetService.listDataSetAccessPermissions(validatedRequest));
     })
   );
 
--- a/solutions/swb-reference/integration-tests/support/resources/datasets/dataset.ts
+++ b/solutions/swb-reference/integration-tests/support/resources/datasets/dataset.ts
@@ -34,6 +34,10 @@ export default class Dataset extends Resource {
     return this._axiosInstance.delete(`/projects/${projectId}/datasets/${this._id}/relationships`);
   }
 
+  public async listAccessPermissions(): Promise<AxiosResponse> {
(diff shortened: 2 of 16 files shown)


bhudnell (approved):

solutions/swb-app/src/dataSetRoutes.ts
```diff
@@ -181,6 +185,21 @@ export function setUpDSRoutes(router: Router, dataSetService: DataSetPlugin): vo
     })
   );
 
+  router.get(
+    '/datasets/:datasetId/permissions',
+    wrapAsync(async (req: Request, res: Response) => {
+      const validatedRequest = validateAndParse<ListDataSetAccessPermissionsRequest>(
+        ListDataSetAccessPermissionsRequestParser,
+        {
+          ...req.query,
```
maghirardelli, line 194: is it safe just to spread the query like this? Looking at the parser, you want pagination token from the query. Why not just specify that clearly?

solutions/swb-reference/integration-tests/tests/multiStep/dataset.test.ts
```diff
@@ -132,6 +133,28 @@ describe('multiStep dataset integration test', () => {
       .dataset(dataSet.id)
       .associateWithProject(unassociatedProject.id, 'read-only');
 
+    console.log('CHECK PROJECT PERMISSIONS FOR DATASET');
+    const { data: responseData } = await adminSession.resources.datasets
+      .dataset(dataSet.id)
+      .listAccessPermissions();
+    const sortedActual: DataSetPermission[] = responseData.data.permissions.sort(
```
maghirardelli, line 140: we love a custom sort method!

solutions/swb-reference/integration-tests/support/resources/datasets/dataset.ts
```diff
@@ -34,6 +34,9 @@ export default class Dataset extends Resource {
     return this._axiosInstance.delete(`/projects/${projectId}/datasets/${this._id}/relationships`);
   }
 
+  public async listAccessPermissions(): Promise<AxiosResponse> {
+    return this._axiosInstance.get(`/datasets/${this._id}/permissions`);
```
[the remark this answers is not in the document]
aws-tyler: This may or may not need to eventually be nested under `projects` for AuthZ, and as far as i can tell, the framework isn't set up to handle changes in paths very easily. Kevin and I have been talking about just getting code in for feature freeze and then evaluating / updating this thing so it better fits our needs. I can change it around for now though if you feel strongly about it.

solutions/swb-reference/integration-tests/tests/multiStep/dataset.test.ts
aws-tyler: i bet theres some backstory here i don't know about haha

solutions/swb-app/src/dataSetRoutes.ts
aws-tyler: The parser will only take the data that it has been instructed to parse. Right now, workbench core's implementation doesn't take a page size, but I think its going to be refactored to do that. Doing it this way means I only have to make a change in the parser if we decide to accept page size in the future. I can change it though if you think its worth doing.

maghirardelli: Yup, that makes sense! Good with this!

solutions/swb-reference/integration-tests/support/resources/datasets/dataset.ts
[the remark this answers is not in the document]
maghirardelli: I think in the future, it would be better to maintain usage of `this._api` since there is a reason you put this test building method under this resource, but I won't make you do it now. 

In the case that you are explaining though, I think that's an argument for using `this._api` because we want route changes for resource/resource collections as a whole to require minimal refactor.

solutions/swb-reference/integration-tests/tests/multiStep/dataset.test.ts
maghirardelli: relying on default sort methods result in very frustrating bugs if the default does not behave how you think it behaves! We faced a bug of this nature for public preview and it was harrowing (okay, I'm being a tad dramatic).

maghirardelli (approved):

PR #875, opened 2023-02-20
feat: verify cost center was not deleted before creating project

otlci:
Issue #, if available: GALI-2282

Description of changes:

* new `status` field returned from cost center API with two possible values `AVAILABLE` and `DELETED`
* each cost center starts with `AVAILABLE` and after soft deleting it switches to `DELETED`
* new check in project creation testing for `status === `AVAILABLE`

Checklist:

### DIFF
--- a/solutions/swb-reference/integration-tests/tests/isolated/projects/create.test.ts
+++ b/solutions/swb-reference/integration-tests/tests/isolated/projects/create.test.ts
@@ -157,5 +157,37 @@ describe('Create Project negative tests', () => {
         }
       });
     });
+
+    describe('that was deleted', () => {
+      beforeEach(async () => {
+        const { data: costCenter } = await adminSession.resources.costCenters.create({
+          name: 'project integration test cost center',
+          accountId: setup.getSettings().get('defaultHostingAccountId'),
+          description: 'a test costcenter'
+        });
+        costCenterId = costCenter.id;
+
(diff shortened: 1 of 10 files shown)


workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -50,7 +51,7 @@ export default class CostCenterService {
       await this._dynamoDbService.updateExecuteAndFormat({
         key: buildDynamoDBPkSk(request.id, resourceTypeToKey.costCenter),
         params: {
-          item: { resourceType: `${this._resourceType}_deleted` }
```
kpark277, line 53: You'll want to keep this as well; internally we will still update the resourceType so that this costCenter no longer shows up when listing, but externally we report the status.

solutions/swb-reference/integration-tests/tests/isolated/projects/create.test.ts
```diff
@@ -123,13 +123,13 @@ describe('Create Project negative tests', () => {
         createRequest.costCenterId = 'cc-invalid-cost-center';
       });
 
-      test('it throws 404 error', async () => {
```
kpark277, line 126: Not sure how this ever succeeded if we always threw Boom.badRequest and not Boom.notFound. The `error` below needs to be updated from `Not Found` to `Bad Request`. Is this test working?

kpark277 (approved):

PR #890, opened 2023-02-22
fix: post deployment setup

otlci:
Issue #, if available:

Description of changes:

Fixed premature group creation in `cognitoSetup` which prevents correct population of DB tables.

Checklist:


DaveMadsen (approved):

workbench-core/environments/src/postDeployment/cognitoSetup.ts
```diff
@@ -33,18 +33,7 @@ export default class CognitoSetup {
...
-      if (e instanceof GroupExistsException) {
-        console.log(`IT Admin group already exists in user pool ${USER_POOL_NAME}`);
-      } else {
-        throw e;
-      }
-    }
-
+    // Create Researcher groups in user pool if they do not exist
     try {
       await this.createGroup('Researcher', userPoolId);
```
htztomic, line 38: Do we still need to create the researcher role?

htztomic (approved):
One comment

workbench-core/environments/src/postDeployment/cognitoSetup.ts
otlci: I opted to keep changes minimal to avoid any potential issues downstream  since I don't have all the context however if removal of Researcher role is safe operation I can delete it.

bhudnell (commented):
Have tests been updated to reflect the changes?

otlci:
> Have tests been updated to reflect the changes?

If I understood it correctly postDeployment is not run for integration tests and unit tests are passing (they don't check for creation of specific user).

workbench-core/environments/src/postDeployment/cognitoSetup.ts
kpark277: This group can be removed.

bhudnell:
> > Have tests been updated to reflect the changes?
> 
> If I understood it correctly postDeployment is not run for integration tests and unit tests are passing (they don't check for creation of specific user).

It looks like the tests in `cognitoSetup.test.ts` mention creating groups and adding the user to a group. Even if the tests still pass, probably worth cleaning up to have the tests describe what is actually happening

DaveMadsen:
Please do not remove the unit tests mentioned unless you are going to remove the methods which they test. This change does not remove those methods as of the current iteration.

bhudnell:
> Please do not remove the unit tests mentioned unless you are going to remove the methods which they test. This change does not remove those methods as of the current iteration.

To clarify, I'm advocating for changing the test descriptions. `run: Create new groups, create new user, add user to Admin group` is no longer accurate since groups are no longer being created, nor is the user added to a group.

bhudnell (changes requested):
Unit test descriptions should be changed to reflect that we are no longer creating groups and assigning the user to the admin group.

DaveMadsen:
> 

NO!!!!
The code that allows group creation is still there! We cannot remove the unit tests until we remove that code entirely and that is not part of this change!

bhudnell (approved):

aws-tyler:
Are these test names still accurate? I'm having a hard time understanding what is actually going on in the tests so I'm not quite sure what to be looking for.

maghirardelli (dismissed):

workbench-core/environments/src/postDeployment/cognitoSetup.test.ts
```diff
@@ -47,41 +37,18 @@ describe('CognitoSetup', () => {
       });
     }
 
-    test('run: Create new groups, create new user, add user to Admin group', async () => {
+    test('run: Create new user', async () => {
```
aws-tyler, line 40: Can name this something like 'adminCreateUser is called during admin creation'? Only suggesting that because the test explicitly is checking if the method is called and whether the `run` method throws an error or not; it doesn't actually check if a user was created.

workbench-core/environments/src/postDeployment/cognitoSetup.test.ts
```diff
@@ -96,30 +63,7 @@ describe('CognitoSetup', () => {
...
-        new UsernameExistsException({
-          $metadata: responseMetadata,
-          message: ''
-        })
-      );
-      jest.spyOn(cognitoSetup, 'createGroup').mockImplementation();
-      jest.spyOn(cognitoSetup, 'adminCreateUser').mockImplementation();
-      const returnVal = await cognitoSetup.run();
-      expect(returnVal).toBeUndefined();
+      expect(cognitoSetup.adminCreateUser).toBeCalledTimes(1);
```
aws-tyler, line 66: It looks like we're expecting the same outputs as the above test even though the user already exists. What are we verifying in this scenario?

DaveMadsen: the response from `AdminCreateUserCommand` has been changed from a `resolve` to a `reject`. this is running the error path.

aws-tyler (approved):

PR #901, opened 2023-02-24
fix(integ tests): clean up all resources created by successful tests

maghirardelli:
Issue #, if available:
GALI-2289

Description of changes:
Our integ test setup is done in a way that the cleanup for the entire test suite should happen exactly once or resources may get left behind without being cleaned up. When writing a new integ test suite, call `.cleanup()` in an `afterAll` and not an `afterEach` block.

This PR also removes the `jest-circus` dependency since we want vanilla jest for consistent behavior throughout test suites.

The integ test suites for SWB should take approximately 40 minutes (as opposed to ~27 minutes when cleanup was not being done properly).

Checklist:

### DIFF
--- a/common/changes/@aws/workbench-core-environments/maghirardelli-project-test-cleanup_2023-02-24-14-52.json
+++ b/common/changes/@aws/workbench-core-environments/maghirardelli-project-test-cleanup_2023-02-24-14-52.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@aws/workbench-core-environments",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@aws/workbench-core-environments"
+}


solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
```diff
@@ -48,6 +48,7 @@ describe('multiStep costCenter test', () => {
     expect(updatedCostCenterB).toMatchObject({ name, description });
 
     console.log('Delete Cost Center B');
-    await expect(adminSession.resources.costCenters.costCenter(createdCostCenterB.id).delete()).resolves;
+    // eslint-disable-next-line no-unused-expressions
+    expect(await adminSession.resources.costCenters.costCenter(createdCostCenterB.id).delete()).resolves;
```
nguyen102, line 52: Are we sure the `await` is supposed to be in the expect?  The example shown [here](https://jestjs.io/docs/asynchronous) shows the format of the original code. 

```
test('the data is peanut butter', async () => {
  await expect(fetchData()).resolves.toBe('peanut butter');
});
```

solutions/swb-reference/integration-tests/tests/isolated/sshKeys/create.test.ts
```diff
@@ -13,34 +13,37 @@ describe('cannot create SSH key', () => {
...
-  afterEach(async () => {
-    await setup.cleanup();
+    costCenterId = costCenter.id;
   });
 
   describe('when the user already has a key for that project', () => {
     let existingSshKeyId: string;
 
     beforeEach(async () => {
+      const { data } = await adminSession.resources.projects.create({
```
nguyen102, line 41: Nitpick: Since there's only one test in the `describe` block, I would just put the `beforeEach` code inside of the `it throws 400 error` test.

If there were multiple tests later on, and the developer thinks each test needs the same setup, they can move that code to `beforeEach`.

solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
maghirardelli: When I ran it the example way, I got one of those "a process is still open you're test might not have finished" jest errors sometimes. I am sure the await outside works while the await inside might cause async issues. I think it might be because the check is just that is resolves, not the resolution value like you have in the example maybe?

solutions/swb-reference/integration-tests/tests/isolated/sshKeys/create.test.ts
maghirardelli: I see what you mean. I think this test got refactored in the PaaB branch anyway so I won't change anything here.

aws-tyler (approved):

solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
DaveMadsen: I recommend

`await expect(adminSession.resources.costCenters.costCenter(createdCostCenterB.id).delete()).resolves.toBeUndefined()`

DaveMadsen (approved):
I added to the conversation with Tim on `await expect()` vs expect(await)`, but if you aren't making changes, I'm not going to block on that. If you do make changes, feel free to ping me for another look!

solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
maghirardelli: We'll look to do that in the future!

PR #918, opened 2023-02-28
feat: multiple app client ids

otlci:
Issue #, if available: MAFoundation-673

Description of changes:

* adding support for passing additional appClientIds

Checklist:


workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -87,13 +80,15 @@ export interface WorkbenchUserPoolOidcIdentityProvider
 
 export class WorkbenchCognito extends Construct {
   public readonly userPool: UserPool;
-  public readonly userPoolClient: UserPoolClient;
+  public readonly webUiUserPoolClient: UserPoolClient;
+  public readonly integrationTestUserPoolClient: UserPoolClient;
```
bhudnell, line 84: Not necessarily for integration tests. Maybe something more generic like "programmaticAccessUserPoolClient"

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -137,43 +132,47 @@ export class WorkbenchCognito extends Construct {
...
-          UserPoolId: this.userPool.userPoolId,
-          ClientId: this.userPoolClient.userPoolClientId
-        },
-        physicalResourceId: PhysicalResourceId.of(this.userPoolClient.userPoolClientId)
-      },
-      policy: AwsCustomResourcePolicy.fromSdkCalls({
-        resources: [this.userPool.userPoolArn]
-      }),
-      installLatestAwsSdk: true
+      userPoolClientName: `${userPoolClientName}-webUi`
```
bhudnell, line 149: We could have a "userPoolWebUiClientName" variable to make this more configurable.

workbench-core/infrastructure/src/workbenchCognito.test.ts
```diff
@@ -165,7 +190,35 @@ describe('WorkbenchCognito tests', () => {
...
+      AccessTokenValidity: 5,
+      RefreshTokenValidity: 1440,
+      TokenValidityUnits: {
+        IdToken: 'minutes',
+        AccessToken: 'minutes',
+        RefreshToken: 'minutes'
+      },
+      ExplicitAuthFlows: [
+        'ALLOW_USER_PASSWORD_AUTH',
+        'ALLOW_ADMIN_USER_PASSWORD_AUTH',
```
bhudnell, line 214: should the web UI client have `ALLOW_ADMIN_USER_PASSWORD_AUTH`?

workbench-core/example/infrastructure/integration-tests/tests/authentication/routeHandler.test.ts
```diff
@@ -110,7 +110,13 @@ describe('authentication route handler integration tests', () => {
...
   describe('refreshAccessToken', () => {
     it('should return the id token if the access token is successfully refreshed', async () => {
-      const adminSession = await setup.createAdminSession();
+      const adminSession = await setup.createRootUserSession(
+        setup.getSettings().get('ExampleCognitoUserPoolId'),
+        setup.getSettings().get('ExampleCognitoWebUiUserPoolClientId'),
+        'USER'
+      ); // Log in using ALLOW_USER_PASSWORD_AUTH flow because ALLOW_ADMIN_USER_PASSWORD_AUTH is disabled for WebUI appClient
+
+      //const adminSession = await setup.createAdminSession();
```
bhudnell, line 119: still needed?

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,6 +29,18 @@ interface TokensExpiration {
   refreshToken: number; // ms
 }
 
+export interface CognitoWebUiClientAuthenticationOptions {
```
bhudnell, line 32: maybe just "CognitoWebUiClient"

workbench-core/authentication/README.md
```diff
@@ -20,8 +20,11 @@ An authentication service that implements the [authorization code grant](https:/
 const cognitoAuthenticationPluginOptions: CognitoAuthenticationPluginOptions = {
   cognitoDomain: '<Cognito Hosted UI Domain>',
   userPoolId: '<Cognito User Pool ID>',
-  clientId: '<Cognito User Pool Client ID>',
-  clientSecret: '<Cognito User Pool Client Secret>',
+  webUiClient: {
+    clientId: '<Cognito User Pool Client ID for WebUI>',
+    clientSecret: '<Cognito User Pool Client Secret for WebUI>',
+  },
+  allowedClientIds: ['<Optional Cognito User Pool ID for integration tests>'],
```
bhudnell, line 27: not necessarily for integration tests. Can be for any programmatic access.

bhudnell (changes requested):
Only comment is that the changes arent explicitly for integration tests, so we'll want to change the variable/doc names to reflect that the new cognito clients are for any programmatic access

workbench-core/infrastructure/src/workbenchCognito.test.ts
DaveMadsen: I thought we discussed this on Slack. No, we shouldn't, but we need it until E2E tests are able to verify the settings. Should be removed once the E2E tests are set up.

bhudnell: `ALLOW_ADMIN_USER_PASSWORD_AUTH` is not required for the web client due to `ALLOW_USER_PASSWORD_AUTH` allowing for the same tests.

otlci: There are two tests in this file, one is `correctly uses default values` and this one. In first one (line 89) it correctly populates the array with only 
```
ExplicitAuthFlows: [
        'ALLOW_USER_PASSWORD_AUTH',
        'ALLOW_CUSTOM_AUTH',
        'ALLOW_USER_SRP_AUTH',
        'ALLOW_REFRESH_TOKEN_AUTH'
      ]
```
but in this second one it generates also `ALLOW_ADMIN_USER_PASSWORD_AUTH`. I assumed it's because additional properties are being passed into the function but I'd like to hear your opinion.

bhudnell: I didnt see any parameters that would allow `ALLOW_ADMIN_USER_PASSWORD_AUTH`, which is why I asked.

otlci: From the way it was written I didn't realize that lodash [merge](https://lodash.com/docs/4.17.15#merge) modifies first parameter so any change is actually propagated to default values which I think is incorrect behavior so I changed both this and `userPoolProps` to first make a deep copy of object to preserve default values. 
Searching through the codebase it looks like `merge` is also being used in `workbench-core/audit/src/plugins/baseAuditPlugin.ts` which will modify data in parameter `auditEntry` - this might be something that consumers are not expecting (@htztomic )

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -105,7 +113,7 @@ export class WorkbenchCognito extends Construct {
       removalPolicy: props.removalPolicy
     };
 
-    const userPoolProps = merge(userPoolDefaults, tempUserPoolProps);
+    const userPoolProps = merge(cloneDeep(userPoolDefaults), tempUserPoolProps);
```
bhudnell, line 116: I think a better solution would be to do `const userPoolProps = merge({}, userPoolDefaults, tempUserPoolProps);` so we can avoid a potentially expensive `cloneDeep`

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -137,43 +140,38 @@ export class WorkbenchCognito extends Construct {
...
-        resources: [this.userPool.userPoolArn]
-      }),
-      installLatestAwsSdk: true
+      userPoolClientName: webUiUserPoolClientName
     });
 
-    const userPoolClientSecret = describeCognitoUserPoolClient.getResponseField(
-      'UserPoolClient.ClientSecret'
-    );
+    const tempIntegrationTestUserPoolClientProps: UserPoolClientOptions = {
```
bhudnell, line 155: change to `programmaticAccess`

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -137,43 +140,38 @@ export class WorkbenchCognito extends Construct {
...
+      authFlows: {
+        adminUserPassword: true
+      }
```
bhudnell, line 158: we should also include separate access/refresh/id token validity fields here and in the `WorkbenchCognitoProps` object so the user can have different lengths of time per client

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -137,43 +140,38 @@ export class WorkbenchCognito extends Construct {
...
+    };
+    merge(userPoolClientProps, tempIntegrationTestUserPoolClientProps);
+    this.programmaticAccessUserPoolClient = new UserPoolClient(this, 'WorkbenchUserPoolClient-iTest', {
+      ...userPoolClientProps,
+      userPool: this.userPool,
+      userPoolClientName: programmaticAccessUserPoolName
+    });
+    this.userPool.identityProviders.forEach((provider) => {
+      this.webUiUserPoolClient.node.addDependency(provider);
+      this.programmaticAccessUserPoolClient.node.addDependency(provider);
```
bhudnell, line 168: I'm not 100% sure but I think the identity providers will only work through the hosted UI, so not necessary for the programmatic access client. I would double check though.

otlci: From what I'm reading here that's correct so I'm going to remove it.

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -135,38 +141,42 @@ export class WorkbenchCognito extends Construct {
...
-    const userPoolClientProps = merge(cloneDeep(userPoolClientDefaults), tempUserPoolClientProps);
+
+    const webUiUserPoolClientProps = merge({}, userPoolClientDefaults, baseUserPoolClientProps);
     this.webUiUserPoolClient = new UserPoolClient(this, 'WorkbenchUserPoolClient-webUi', {
-      ...userPoolClientProps,
+      ...webUiUserPoolClientProps,
       userPool: this.userPool,
-      userPoolClientName: webUiUserPoolClientName
+      userPoolClientName: webUiUserPoolClientName,
+      ...webUiUserPoolTokenValidity
```
bhudnell, line 157: need to merge this with the defaults

workbench-core/infrastructure/src/workbenchCognito.ts
```diff
@@ -135,38 +141,42 @@ export class WorkbenchCognito extends Construct {
...
-      ...userPoolClientProps,
+      ...programmaticAccessUserPoolClientProps,
       userPool: this.userPool,
-      userPoolClientName: programmaticAccessUserPoolName
-    });
-    this.userPool.identityProviders.forEach((provider) => {
-      this.webUiUserPoolClient.node.addDependency(provider);
-      this.programmaticAccessUserPoolClient.node.addDependency(provider);
+      userPoolClientName: programmaticAccessUserPoolName,
+      ...programmaticAccessUserPoolTokenValidity
```
bhudnell, line 175: need to merge this with the defaults

solutions/swb-reference/src/SWBStack.ts
```diff
@@ -1469,9 +1469,13 @@ export class SWBStack extends Stack {
       domainPrefix: domainPrefix,
       websiteUrls: websiteUrls,
       userPoolName: userPoolName,
-      userPoolClientName: userPoolClientName,
+      webUiUserPoolClientName: `${userPoolClientName}-webUi`,
+      programmaticAccessUserPoolName: `${userPoolClientName}-iTest`,
       oidcIdentityProviders: [],
-      accessTokenValidity: Duration.minutes(60) // Extend access token expiration to 60 minutes to allow integration tests to run successfully. Once MAFoundation-310 has been implemented to allow multiple clientIds, we'll create a separate client for integration tests and the "main" client access token expiration time can be return to 15 minutes
```
bhudnell, line 1474: Have you double checked that SWB integ tests will all pass with this change? Since we're removing the admin auth from the web UI client, I dont think their current workflow will work.

workbench-core/example/infrastructure/src/example-stack.ts
```diff
@@ -708,9 +710,16 @@ export class ExampleStack extends Stack {
...
       websiteUrls: websiteUrls,
       userPoolName: userPoolName,
-      userPoolClientName: userPoolClientName,
+      webUiUserPoolClientName: `${userPoolClientName}-webUi`,
+      programmaticAccessUserPoolName: `${userPoolClientName}-iTest`,
       oidcIdentityProviders: [],
-      accessTokenValidity: Duration.minutes(60), // Extend access token expiration to 60 minutes to allow integration tests to run successfully. Once MAFoundation-310 has been implemented to allow multiple clientIds, we'll create a separate client for integration tests and the "main" client access token expiration time can be return to 15 minutes
+      // Extend access token expiration to 60 minutes to allow integration tests to run successfully. Once MAFoundation-310 has been implemented to allow multiple clientIds, we'll create a separate client for integration tests and the "main" client access token expiration time can be return to 15 minutes
+      webUiUserPoolTokenValidity: {
+        accessTokenValidity: Duration.minutes(60)
```
bhudnell, line 718: Can we revert this back to the default value? Or does it have to wait until we have e2e tests to fully step away from using the webUi cleint for integ tests?

otlci: I kept this value since I added one test that uses WebUI token (`routeHandler.test.refreshAccessToken`). My understanding is that we can remove this once we have E2E tests.

solutions/swb-reference/src/SWBStack.ts
otlci: Good point. I'm going to rebuild whole SWB environment from scratch and retest.

otlci: I'm not 100% how SWB app clients are being used so I opted to duplicate setup in `workbench-core` integration tests (both WebUI and programmatic access app clients are allowed).

bhudnell (approved):

workbench-core/authentication/src/plugins/cognitoAuthenticationPlugin.ts
```diff
@@ -29,6 +29,18 @@ interface TokensExpiration {
   refreshToken: number; // ms
 }
 
+export interface CognitoWebUiClient {
```
kpark277, line 32: Rather than setting up a separate client, could we just allow the end user to decide if they want to use the web ui client vs. programmatic client, they can pass those in when setting up the plugin?

kpark277 (dismissed):

kpark277 (dismissed):

bhudnell (dismissed):

kpark277 (approved):

bhudnell (approved):

PR #919, opened 2023-02-28
chore: use Zod in Create Cost Center API

aws-tyler:
Adding zod parsing to the request for Create Cost Center API


workbench-core/accounts/src/models/costCenters/createCostCenterRequest.ts
```diff
@@ -3,10 +3,13 @@
...
-interface CreateCostCenterRequest {
-  name: string;
-  accountId: string;
-  description: string;
-}
+import { z } from 'zod';
 
-export default CreateCostCenterRequest;
+// eslint-disable-next-line @rushstack/typedef-var
+export const CreateCostCenterRequestParser = z.object({
```
FernandoAranda, line 9: should this be strict?

aws-tyler: I don't think it needs to be? If i remember right strict causes the parse to fail if extra fields are there, but since the parser is only allowing known parameters through then i don't think we necessarily need to mark it as strict.

FernandoAranda (approved):

PR #935, opened 2023-03-09
chore: Add Zod validation to user APIs

FernandoAranda:
Issue #, if available:
GALI-2090
Description of changes:
Add Zod validation to users APIs inputs

Checklist:


DaveMadsen (approved):
Changes look good. Were you able to run the MAF integration tests?

solutions/swb-app/src/users/updateUserRequest.ts
```diff
@@ -8,6 +8,7 @@ import { z } from 'zod';
 // eslint-disable-next-line @rushstack/typedef-var
 export const UpdateUserRequestParser = z
   .object({
+    userId: z.string().min(1),
     firstName: z.string().optional(),
```
aws-tyler, line 12: do these also need mins?

aws-tyler (approved):
We should revisit these when we have time to make sure all the lengths are optimal, but I think this works well for now.

solutions/swb-app/src/users/updateUserRequest.ts
FernandoAranda: this is a required value and there is no static route config to avoid empties, so adding the regex later may get rid of this

FernandoAranda: this is going to be addressed in the validation input tasks

FernandoAranda:
> Changes look good. Were you able to run the MAF integration tests?

I did run integration tests for SWB, I took a look at the integ test for the example app and they do not specify message, the error code is still the same

PR #949, opened 2023-03-15
fix: remove AuthenticatedUser from datasets DDB entries

bhudnell:
Issue #, if available:

Description of changes:
- Removed `AuthenticatedUser` from the Datasets DDB entries

Checklist:


htztomic (changes requested):
add testing for addDataSet/other dependencies of `_storeEndpointToDdb` and `_storeDataSetToDdb` to check authenticatedUser is not being stored,  should also omit on `provisionDataSet` in the instance that a customer's custom DataSetMetadataPlugin does not omit.

htztomic (approved):
LGTM

PR #988, opened 2023-04-05
DO NOT MERGE: SWBv2 Core Review PR

aws-tyler:
DO NOT MERGE THIS PR. IT IS ONLY BEING USED FOR CODE REVIEW PURPOSES AS A PART OF THE PUBLISHER REVIEW.


buildspec.yml
```diff
@@ -0,0 +1,42 @@
+version: 0.2
+phases:
+  install:
+    runtime-versions:
+      nodejs: 14
```
tbelmega, line 5: Should we upgrade nodejs to the latest avaible version of the build environment?

PR #1065, opened 2023-05-15
fix: remove deployment of example app from pipeline

htztomic:
Issue #, if available:

Description of changes:

Checklist:

### DIFF
--- a/.github/workflows/integration-tests-release.yml
+++ b/.github/workflows/integration-tests-release.yml
@@ -18,16 +18,6 @@ jobs:
           poll-interval-seconds: 10
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  deploy-example-app:
-    needs: [pre-deployment-check]
-    uses: ./.github/workflows/deploy-integration-example-app.yml
-    with:
-      environment: release
-    secrets:
-      aws-region: ${{ secrets.AWS_RELEASE_REGION }}
-      role-to-assume: ${{ secrets.RELEASE_ASSUME_ROLE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests-release.yml
```diff
@@ -18,16 +18,6 @@ jobs:
           poll-interval-seconds: 10
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  deploy-example-app:
```
insignias, line 21: Comment it instead, may be we want to add it back later ?

.github/workflows/merge-develop-to-stage.yml
```diff
@@ -84,22 +84,10 @@ jobs:
       slack_webhook_url: ${{ secrets.SLACK_WEBHOOK }}
       HOSTED_ZONE_ID: ${{ secrets.HOSTED_ZONE_ID_DEMO }}
       DOMAIN_NAME: ${{ secrets.DOMAIN_NAME_DEMO }}
-
-  deploy-example-app:
```
maghirardelli, line 88: maybe just remove (I agree with Samir to actually comment not delete) in the release workflow. The develop workflow can be refactored in the later ticket to move wbc to rsw. so don't edit this file _yet_

.github/workflows/integration-tests-release.yml
maghirardelli: agree

maghirardelli: You may actually be able to skip by adding this line to the step:
```
if: false
```

htztomic: I thought the current ticket is to completely remove example and once wbc is moved into swb, is there any reason to enable example app again?

PR #1069, opened 2023-05-15
fix(environment): failed environment edge cases fixes

nguyen102:
Issue #, if available:

Description of changes:
* Starting/stopping a failed environment should return 409 error
* Allow users to terminate failed environment
* Added integ test for starting failed environment. 
* Added `ts-jest` as a dev dependency so users can run integ test in their IDE

Checklist:


solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
```diff
@@ -31,6 +34,27 @@ describe('environment start negative tests', () => {
...
+      const failedEnvId = settings.get('failedEnvId');
+      const defaultProjId = settings.get('projectId');
+      try {
+        await itAdminSession.resources.projects
+          .project(defaultProjId)
+          .environments()
+          .environment(failedEnvId)
+          .start();
+      } catch (e) {
+        console.log('e', e);
```
SanketD92, line 47: Is the console.log needed?

solutions/swb-app/src/projectEnvironmentRoutes.ts
```diff
@@ -146,6 +146,8 @@ export function setUpProjectEnvRoutes(
       const envType = environment.ETC!.type;
       if (environment.status === 'STOPPING') {
         throw Boom.conflict('Cannot start environment while environment is currently being stopped');
+      } else if (['FAILED'].includes(environment.status)) {
```
janeyuaws, line 149: just use: environment.status === 'FAILED'

janeyuaws (commented):
Hi Tim, I'd suggest you take a look at paabHelper, it helps create PaabResources (projects for PA1, PA2 etc.). Avoid creating new project in the test env. In this case you can test agaist pa1Session and use project1Id.

solutions/swb-app/src/projectEnvironmentRoutes.ts
janeyuaws: can you double check if this error check should only happen at route level or service level. I'd lean towards the later

solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
```diff
@@ -31,6 +34,27 @@ describe('environment start negative tests', () => {
   });
 
   describe('ITAdmin tests', () => {
+    test('environment in failed state', async () => {
+      const failedEnvId = settings.get('failedEnvId');
+      const defaultProjId = settings.get('projectId');
+      try {
+        await itAdminSession.resources.projects
+          .project(defaultProjId)
```
janeyuaws, line 42: I'd suggest you take a look at paabHelper. it helps create PaabResources (projects for PA1, PA2 etc.). Avoid creating new project in the test env. In this case you can test against pa1Session and use project1Id.

solutions/swb-reference/integration-tests/support/utils/settings.ts
```diff
@@ -21,6 +22,7 @@ interface Setting {
   projectAdmin2PasswordParamStorePath: string;
   costCenterId: string;
   projectName: string;
+  failedEnvId: string;
```
janeyuaws, line 25: if error check happens at service level, you can avoid adding this here

solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
nguyen102: good catch, deleting

solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
nguyen102: Ah yes, I looked at `paabHelper` and I see that it create new projects. However, in this test, we can't use preexisting projects because we need an environment in the `failed` state. 

I don't think there's a straight forward programmatic way to create an environment that is in the `FAILED` state, so I went with the option that the dev should preset an environment in the failed state. I saw that we are already requiring devs to set up a default project in the config file. This code will now read in the default project value.

Let me know if this is ok.

solutions/swb-app/src/projectEnvironmentRoutes.ts
nguyen102: Updated the conditional to be ` else if (environment.status === 'FAILED')` since we're only checking one string value.

nguyen102: I think the error check on the route level makes sense because we'll want the same behavior for all environments. If we add the check at the service level, we will have to reuse the code for each new environment type. Also there's a precedence for having these types of checks on the route level. Examples [here](https://github.com/aws-solutions/research-service-workbench-on-aws/blob/release/4debe38e-5796-47b6-96ec-881a87e0df21/solutions/swb-app/src/projectEnvironmentRoutes.ts#L149)

@SanketD92 , @maghirardelli , @kpark277 , @aws-tyler thoughts?

solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
nguyen102: @SanketD92 , @maghirardelli , @kpark277 , @aws-tyler thoughts?

solutions/swb-app/src/projectEnvironmentRoutes.ts
kpark277: Having the check at the service level as well would be great, but I think we can skip for now to minimize how much code we're changing for launch. Only nit is to combine this with the previous line and instead make the error message `Cannot start environment while environment is in ${env.status} state`

solutions/swb-reference/integration-tests/tests/isolated/environments/start.test.ts
nguyen102: I've move the tests to multistep env test, and used paabHelper

solutions/swb-app/src/projectEnvironmentRoutes.ts
nguyen102: keeping error checks on the route level for now

solutions/swb-reference/integration-tests/tests/multiStep/environment.test.ts
```diff
@@ -70,27 +65,37 @@ describe('multiStep environment test', () => {
...
+    const adminSessionEnvironments: Environments = adminSession.resources.projects
+      .project(projectId)
+      .environments();
+    const paSessionEnvironments: Environments = paSession.resources.projects
+      .project(projectId)
+      .environments();
+
+    console.log('Launch Environment A');
+    const environmentA = await launchEnvironment(
+      randomTextGenerator.getFakeText('environment-multistep-test-envA'),
```
SanketD92, line 87: nit: random text generation can be moved inside the launchEnvironment function.

SanketD92 (dismissed):

solutions/swb-reference/integration-tests/tests/multiStep/environment.test.ts
```diff
@@ -126,7 +130,49 @@ describe('multiStep environment test', () => {
     });
     console.log('Environment B Completed');
 
-    //Verify Connect Environment A
+    console.log('Set Environment B state to FAILED');
```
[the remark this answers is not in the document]
nguyen102: Ah, so part of this task is to allow terminating failed environments. Therefore, terminating the env will still succeed.

maghirardelli (approved):

SanketD92 (approved):

solutions/swb-reference/integration-tests/tests/multiStep/environment.test.ts
nguyen102: fixed

SanketD92 (approved):

PR #1074, opened 2023-05-17
chore: update codecove thresholds 

Bingjiling:
Description of changes:
* 4 codecove checks are failing for PRs for not reaching the defined thresholds 
* The thresholds defined in codecove is higher than config/jest file used for build
* Update the 4 failed checks thresholds to be in sync with config/jest file 

Checklist:

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,7 +10,7 @@ coverage:
         flags:
           - example-ui-app
       swb-app:
-        target: 90%
+        target: 70%
         flags:
           - swb-app
       swb-reference:
@@ -34,11 +34,11 @@ coverage:
         flags:
           - workbench-core-authentication
       workbench-core-authorization:
-        target: 100%
+        target: 97%
         flags:
           - workbench-core-authorization
       workbench-core-base:
-        target: 90%
+        target: 70%
         flags:
           - workbench-core-base
       workbench-core-datasets:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -34,11 +34,11 @@ coverage:
         flags:
           - workbench-core-authentication
       workbench-core-authorization:
-        target: 100%
+        target: 97%
         flags:
           - workbench-core-authorization
       workbench-core-base:
-        target: 90%
+        target: 70%
```
SanketD92, line 41: Currently it's at 99.52% right? Can we update this value to 95%?

nguyen102: Hmm, maybe make this 95 and update the jest config too.

nguyen102:
Should we update the jest config to the highest test coverage value that we currently have coverage for, and then move the code cov coverage down to match that level? This will keep our coverage as high as possible and allow us to pass the code cov checks.

codecov.yml
Bingjiling: workbench-core-base is at 81.13%, according to codecov.

Bingjiling:
> Should we update the jest config to the highest test coverage value that we currently have coverage for, and then move the code cov coverage down to match that level? This will keep our coverage as high as possible and allow us to pass the code cov checks.

Yea, I can adjust the changes that was lowered to 70% up to 75% as they're currently at ~80%. This will leave some wiggle room so the checks are not over sensitive. Thoughts?

maghirardelli (approved):

PR #1081, opened 2023-05-18
fix(account-status): working account status update for byon

nguyen102:
Issue #, if available:

Description of changes:
* Update account handler to check `onboard-account-byon.cfn.yml`, `onboard-account-tgw.cfn.yml`, and `onboard-account.cfn.yml` so if users have a hosting account that isn't `onboard-account.cfn.yml`, their account status is correct.

Testing
* This works with `onboard-account.cfn.yml` and `onboard-account-byon.cfn.yml`. I didn't test it with `onboard-account-tgw.cfn.yml`, but it follows the same logic as BYON and should work as expected. 

Checklist:

*  [x] Have you successfully deployed to an AWS account with your changes?

### DIFF
--- a/workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
+++ b/workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
@@ -396,13 +396,12 @@ export default class HostingAccountLifecycleService {
         return output.OutputKey === 'EncryptionKeyArn';
       })!.OutputValue;
 
-      if (
-        expectedTemplates
-          .map((template) => {
-            return removeCommentsAndSpaces(template);
-          })
-          .includes(removeCommentsAndSpaces(actualTemplate))
-      ) {
+      const expectedTemplatesWithoutCommentsAndSpaces = expectedTemplates.map((template) => {
+        return removeCommentsAndSpaces(template);
+      });
+
+      // If the actual template matches one of the expected template then the account is current
+      if (expectedTemplatesWithoutCommentsAndSpaces.includes(removeCommentsAndSpaces(actualTemplate))) {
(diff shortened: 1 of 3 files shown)


workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -385,7 +396,13 @@ export default class HostingAccountLifecycleService {
         return output.OutputKey === 'EncryptionKeyArn';
       })!.OutputValue;
 
-      if (removeCommentsAndSpaces(actualTemplate) === removeCommentsAndSpaces(expectedTemplate)) {
+      if (
+        expectedTemplates
```
aws-tyler, line 400: lets pull this out into its own value so its easier to read.

aws-tyler (dismissed):

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
SanketD92: +1, also need some comments I think. 

How is this piece of code determining which template to match against? Let's say there has been a big change in all 3 expected templates, then how would this work?

nguyen102: Ah yes, I'll add comments and break the code out to make it easier to understand. The logic is the following. If the cfn template for an account matches one of the templates in S3 we'll consider the account is up to date. If a big change was made to all 3 expected templates, then the CFN template won't match any of the 3 expected templates and the account's status will change to `NEEDS_UPDATE`.


This keep the logic simple. However users need to be careful that when they manually onboard a hosting account, they choose the right CFN template.

PR #1083, opened 2023-05-19
feat: Project / Dataset association check

aws-tyler:
Issue #, if available:

Description of changes:
* Updated code to check if a project and dataset are associated before trying to associate them. 
* Updated how DatabaseService retrieves existing associations.
* Added unit and integration tests.

Checklist:


solutions/swb-app/src/projectEnvTypeConfigs/projectEnvTypeConfigPlugin.ts
```diff
@@ -44,7 +44,7 @@ export interface ProjectEnvTypeConfigPlugin {
    * @param request - Request Object to get Environment Type Config
    *
    */
-  getEnvTypeConfig(request: GetProjectEnvTypeConfigRequest): Promise<EnvironmentTypeConfig>;
+  getEnvTypeConfig(request: GetProjectEnvTypeConfigRequest): Promise<EnvironmentTypeConfig | undefined>;
```
maghirardelli, line 47: Why return undefined and not just throw within this method?

maghirardelli (approved):
non blocking comment

solutions/swb-app/src/projectEnvTypeConfigs/projectEnvTypeConfigPlugin.ts
aws-tyler: Its a style choice. IMO it simplifies the code and makes it more readable. The short version is that its good to explicitly call out failure scenarios instead of relying to knowing when to use a `try/catch` block based on knowing when a method will throw an error.

PR #1103, opened 2023-06-01
feat: list users by role pagination

aws-tyler:
Issue #, if available:

Description of changes:
Added pagination to the listUsersByRole endpoint

Checklist:

### DIFF
--- a/workbench-core/user-management/src/plugins/cognitoUserManagementPlugin.ts
+++ b/workbench-core/user-management/src/plugins/cognitoUserManagementPlugin.ts
@@ -11,7 +11,6 @@ import {
   ListUsersForRoleRequest,
   PaginatedResponse
 } from '@aws/workbench-core-base';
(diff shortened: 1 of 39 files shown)


workbench-core/user-management/src/plugins/cognitoUserManagementPlugin.ts
```diff
@@ -491,6 +523,9 @@ export class CognitoUserManagementPlugin implements UserManagementPlugin {
         }
         throw new PluginConfigurationError(error.message);
       }
+      if (error.name === 'InvalidParameterException') {
+        throw new InvalidPaginationTokenError('Invalid Pagination Token');
```
kpark277, line 527: Nit: I imagine Cognito returns this same exception if something like Limit is given an invalid value. Could we just bubble up error.message here?

kpark277 (approved):

PR #1107, opened 2023-06-03
fix: create account should throw 400 error instead of 500 errors for incorrect accountId and arns

nguyen102:
Description of changes:

Add additional Input validation for creating new hosting accounts
* User must provide an `awsAccountId` of a real account
* `envMgmtRoleArn` and `hostingAccountHandlerRole` arn must follow the proper ARN format. 

Checklist:

### DIFF
--- a/solutions/swb-app/src/accountRoutes.ts
+++ b/solutions/swb-app/src/accountRoutes.ts
@@ -15,7 +15,8 @@ import {
   UpdateAccountRequest,
   UpdateAccountRequestParser,
   GetAccountRequest,
-  GetAccountRequestParser
+  GetAccountRequestParser,
+  isInvalidAwsAccountIdError
 } from '@aws/workbench-core-accounts';
 import { isInvalidPaginationTokenError } from '@aws/workbench-core-base';
 import * as Boom from '@hapi/boom';
@@ -69,9 +70,15 @@ export function setUpAccountRoutes(router: Router, hostingAccountService: Hostin
     '/awsAccounts',
     wrapAsync(async (req: Request, res: Response) => {
       const validatedRequest = validateAndParse<CreateAccountRequest>(CreateAccountRequestParser, req.body);
-
-      const createdAccount = await hostingAccountService.create(validatedRequest);
-      res.status(201).send(createdAccount);
+      try {
+        const createdAccount = await hostingAccountService.create(validatedRequest);
+        res.status(201).send(createdAccount);
+      } catch (e) {
+        if (isInvalidAwsAccountIdError(e)) {
+          throw Boom.badRequest(e.message);
+        }
+        throw e;
+      }
     })
   );
 
--- a/workbench-core/accounts/src/errors/errors.test.ts
+++ b/workbench-core/accounts/src/errors/errors.test.ts
@@ -1,0 +1,14 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InvalidAwsAccountIdError, isInvalidAwsAccountIdError } from './InvalidAwsAccountIdError';
+
+describe('custom error tests', () => {
+  test('InvalidAwsAccountIdError', () => {
+    const invalidAwsAccountIdError = new InvalidAwsAccountIdError();
+
+    expect(isInvalidAwsAccountIdError(invalidAwsAccountIdError)).toBe(true);
+  });
+});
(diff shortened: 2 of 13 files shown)


workbench-core/accounts/src/utilities/hostingAccountLifecycleService.test.ts
```diff
@@ -642,6 +644,76 @@ describe('HostingAccountLifecycleService', () => {
...
+        ]
+    }`
+    });
+    const s3ServiceException = new S3ServiceException({
+      $fault: 'client',
+      $metadata: {},
+      name: 'MalformedPolicy'
+    });
+
+    // @ts-ignore
```
aws-tyler, line 705: why do we need the ignore here?

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -203,7 +209,18 @@ export default class HostingAccountLifecycleService {
     };
 
     // Update bucket policy
-    await this._aws.clients.s3.putBucketPolicy(putPolicyParams);
+    try {
+      await this._aws.clients.s3.putBucketPolicy(putPolicyParams);
+    } catch (e) {
+      if (
+        e.name === 'MalformedPolicy' &&
```
aws-tyler, line 216: might want to pull the condition into a variable called `isInvalidAccountId` or something to make it easier to read whats happening here.

solutions/swb-app/src/accountRoutes.ts
```diff
@@ -15,7 +15,8 @@ import {
   UpdateAccountRequest,
   UpdateAccountRequestParser,
   GetAccountRequest,
-  GetAccountRequestParser
+  GetAccountRequestParser,
+  isInvalidAwsAccountIdError
```
kpark277, line 19: Nit: Pascal casing

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
```diff
@@ -651,12 +668,12 @@ export default class HostingAccountLifecycleService {
...
     const { statusHandlerArn, artifactBucketArn, mainAcctEncryptionArnList } = arns;
 
-    // Update main account default event bus to accept hosting account state change events
-    await this.updateBusPermissions(statusHandlerArn, awsAccountId);
-
     // Add account to artifactBucket's bucket policy
     await this.updateArtifactsBucketPolicy(artifactBucketArn, awsAccountId);
 
+    // Update main account default event bus to accept hosting account state change events
+    await this.updateBusPermissions(statusHandlerArn, awsAccountId);
```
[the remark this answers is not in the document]
aws-tyler: i think this is ok for now since we're short on time, but relying on the order seems brittle since its not clear that call order for the functions is critical to things working correctly. would it be possible to know whether the account is valid by the time we get here so we know we're safe? i'm guessing no but thought i'd ask just in case.

solutions/swb-app/src/accountRoutes.ts
aws-tyler: Its a function, not a class, so wouldn't it still be camel case?

kpark277 (approved):

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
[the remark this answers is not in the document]
nguyen102: I agree that this is brittle. There isn't a clear cut way to figure out if the accountId actually belongs to a real aws account. AWS doesn't have an API for that. It only comes up as an error through using other APIs that rely on real account ids. 

Maybe adding a comment about the order will be helpful?

workbench-core/accounts/src/errors/errors.test.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InvalidAwsAccountIdError, isInvalidAwsAccountIdError } from './InvalidAwsAccountIdError';
+
+describe('custom error tests', () => {
+  test('InvalidAwsAccountIdError', () => {
```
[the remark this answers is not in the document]
nguyen102: I was following the pattern here.
https://github.com/aws-solutions/research-service-workbench-on-aws/blob/main/workbench-core/datasets/src/errors/errors.test.ts

However, your suggestion does make sense and would be simple enough to implement.

workbench-core/accounts/src/utilities/hostingAccountLifecycleService.ts
nguyen102: Hmm, would that make it easier to read? I think the if statement with the 3 conditions are pretty self explanatory. I'm glad to make it a variable if you strongly feel it needs to be though.

`if (malformedPolicy && detail is related to arn with awsAccountId && is S3ServiceException) throw error`.

PR #1122, opened 2023-06-09
chore: etc endpoint integ tests

aws-tyler:
Issue #, if available:

Description of changes:
Added positive and negative integration tests for the List and Get endpoints associated with all possible ways of getting Environment Type Configs

Checklist:


solutions/swb-reference/integration-tests/tests/isolated/environmentTypeConfigs/get.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      }
+    });
+
+    test('fails as a Researcher', async () => {
+      try {
+        await researcherSession.resources.environmentTypes
+          .environmentType(envTypeId)
+          .configurations()
+          .environmentTypeConfig(envTypeConfigId)
+          .get();
```
nguyen102, line 69: Do we want to throw an error after the `await` statement to guarantee that the `await` statement threw an error? 

An example is shown [here](https://github.com/aws-solutions/research-service-workbench-on-aws/blob/d035cc34c555b521cc50e5aab88924e5c39c2fa1/solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts#L38). We might have a false negative where the test passes because the `await` statement doesn't throw an error.

solutions/swb-reference/integration-tests/tests/isolated/environmentTypeConfigs/get.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          e,
+          new HttpError(403, {
+            error: 'User is not authorized'
+          })
+        );
+      }
+    });
+  });
+
+  test('get environments type configs (non-project route) fails', async () => {
```
nguyen102, line 81: Is this a duplicate, there's a describe statement on line 35 with the same string as this `get environments type configs (non-project route)`. 

Is this the same test as on line 46?

nguyen102 (commented):
Overall looks good, some minor comments.

solutions/swb-reference/integration-tests/tests/isolated/environmentTypeConfigs/get.test.ts
nguyen102: Actually I don't think this is needed for this scenario since you already have `jest.hasAssertions` and each of these tests only have one assertion.

PR #1126, opened 2023-06-09
fix: separate GetDataset and ListDatasets permissions

kpark277:
Description of changes:
* GetDataset and ListDatasets updated to use different authz permission subjects
* Removed Experimental flag from all published packages
* Updated PaabHelper to allow users to set how many projects they want to create

Checklist:


solutions/swb-reference/integration-tests/support/complex/paabHelper.ts
```diff
@@ -24,14 +24,20 @@ export class PaabHelper {
   private _setup: Setup;
   private _randomTextGenerator: RandomTextGenerator;
   private _outputError: boolean | undefined;
+  private _numberOfProjects: number = 3;
 
-  public constructor(outputError?: boolean) {
+  public constructor(numberOfProjects?: number, outputError?: boolean) {
```
nguyen102, line 29: Does `numberOfProjects` need to be passed into the constructor? We can keep the scope smaller by passing in `numberOfProjects` to `createResources`. That's the only method using it. That way we won't need a class variable for a variable that's only being used in one method.

solutions/swb-reference/integration-tests/support/complex/paabHelper.ts
```diff
@@ -41,20 +47,18 @@ export class PaabHelper {
...
+      const projectName = `Project${i}`;
       const projectResponse = await adminSession.resources.projects.create({
         name: this._randomTextGenerator.getFakeText(projectName),
         description: `${projectName} for integ tests`,
         costCenterId: costCenter.id
       });
       projectIds.push(projectResponse.data.id);
     }
+
     const [project1Id, project2Id, project3Id] = projectIds;
```
nguyen102, line 62: What's the expected behavior if `_numberOfProjects = 1`? Would `project2Id` and `project3Id` be `undefined`?

solutions/swb-reference/integration-tests/support/complex/paabHelper.ts
```diff
@@ -24,14 +24,20 @@ export class PaabHelper {
...
+  public constructor(numberOfProjects?: number, outputError?: boolean) {
     this._setup = Setup.getSetup();
     this._randomTextGenerator = new RandomTextGenerator(this._setup.getSettings().get('runId'));
     this._outputError = outputError;
+    if (numberOfProjects) this._numberOfProjects = numberOfProjects;
   }
 
   public async createResources(): Promise<PaabResources> {
+    if (this._numberOfProjects > 3) {
+      throw new Error('PaabHelper cannot support more than 3 projects');
```
SanketD92, line 38: What is the downside of allowing the creation of all resources every time? Is the cleanup method skipping resource deletion?

solutions/swb-reference/integration-tests/tests/isolated/projects/get.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  afterAll(async () => {
+    await paabHelper.cleanup();
+  });
+
+  describe('when a ProjectAdmin GETs a project they do not belong to', () => {
+    test('it returns a 403', async () => {
+      try {
+        await pa1Session.resources.projects.project(project2Id).get();
```
SanketD92, line 35: Let's also throw an [error](https://github.com/aws-solutions/research-service-workbench-on-aws/pull/1125#discussion_r1223821122) after the API call in case it doesn't throw an error.

solutions/swb-reference/integration-tests/support/complex/paabHelper.ts
kpark277: Yup, they get set to undefined.

solutions/swb-reference/integration-tests/support/complex/paabHelper.ts
kpark277: Yeah, I realized it would have been better on createResources after I had refactored all the tests; I can go back through and update them all if we think it's worth changing, but I don't see a reason to do that now until we come back to make this into a proper singleton.

solutions/swb-reference/integration-tests/tests/isolated/projects/get.test.ts
kpark277: Test fails if it never runs into an expect.

SanketD92 (approved):
LGTM

PR #1133, opened 2023-06-12
feat: CostCenter auth tests

aws-tyler:
Added positive and negative auth tests for all CostCenter endpoints.


solutions/swb-reference/integration-tests/tests/isolated/costCenter/create.test.ts
```diff
@@ -86,28 +129,29 @@ describe('Cost Center negative tests', () => {
...
-              new HttpError(400, {
-                error: 'Bad Request',
-                message: `accountId: Invalid ID`
-              })
-            );
-          }
-        })
-      );
+
+      for (const invalidRequest of invalidCreateRequests) {
```
aws-tyler, line 142: was getting intermittent 502s when using `Promise.all`, so I switched to a for loop which appeared to be more stable.

solutions/swb-reference/integration-tests/tests/isolated/costCenter/softDelete.test.ts
```diff
@@ -1,29 +1,85 @@
...
+    await paabHelper.cleanup();
+  });
+
+  describe('authorization test:', () => {
+    let costCenterId: string;
+    beforeEach(async () => {
+      const accountId = setup.getSettings().get('defaultHostingAccountId');
+      const { data: createdCostCenterA } = await itAdminSession.resources.costCenters.create({
+        accountId,
+        name: 'costCenterA'
```
Bingjiling, line 36: Nit: can we name it something like 'costCenterForIntegTest'?

solutions/swb-reference/integration-tests/tests/isolated/costCenter/update.test.ts
```diff
@@ -45,7 +106,7 @@ describe('Update Cost Center negative tests', () => {
       try {
         const name =
```
Bingjiling, line 107: Can we use repeat() here?

solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
```diff
@@ -16,39 +16,51 @@ describe('multiStep costCenter test', () => {
...
+    const costCenterA = createCostCenterAResponse.data;
+    const costCenterB = createCostCenterBResponse.data;
 
     console.log('Get Cost Center A');
-    const { data: getCostCenterA } = await adminSession.resources.costCenters
-      .costCenter(createdCostCenterA.id)
-      .get();
-    expect(getCostCenterA).toMatchObject(createdCostCenterA);
+    const responseCostCenterA = await itAdminSession.resources.costCenters.costCenter(costCenterA.id).get();
+    expect(responseCostCenterA.status).toEqual(200);
```
Bingjiling, line 36: nit: `expect(responseCostCenterA).toEqual(expect.objectContaining({status: 200, data:costCenterA}))` should work here.

solutions/swb-reference/integration-tests/tests/isolated/costCenter/update.test.ts
aws-tyler: same as above.

solutions/swb-reference/integration-tests/tests/multiStep/costCenter.test.ts
aws-tyler: I think we probably want to avoid compound `expect`s when there are only two things being checked because the error reporting is easier to read, especially if you have responses with large amounts of text in them.

solutions/swb-reference/integration-tests/tests/isolated/costCenter/create.test.ts
aws-tyler: Lets keep it for now since its already present. I'm not sure why Fernando initially did different variations but I think changing that would be outside the scope of this update

solutions/swb-reference/integration-tests/tests/isolated/costCenter/get.test.ts
```diff
@@ -58,7 +116,7 @@ describe('Get Cost Center negative tests', () => {
       const invalidId =
```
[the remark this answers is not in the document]
Bingjiling: It should be a 1min change, but I'm fine leaving it outside now.

Bingjiling (approved):

solutions/swb-reference/integration-tests/tests/isolated/costCenter/create.test.ts
```diff
@@ -26,6 +34,41 @@ describe('Cost Center negative tests', () => {
...
+      expect(response.status).toEqual(201);
+    });
+
+    test('ProjectAdmin cannot create CostCenter', async () => {
+      try {
+        await pa1Session.resources.costCenters.create(validCreateRequest);
+      } catch (e) {
+        checkHttpError(
+          e,
+          new HttpError(403, {
```
SanketD92, line 53: Let's call this error up top in the describe section once and reference it everywhere:
```const unauthorizedHttpError = new HttpError(403, { error: 'User is not authorized' });```

File: solutions/swb-reference/integration-tests/tests/isolated/costCenter/create.test.ts [hunk 7]
```diff
@@ -86,28 +129,29 @@ describe('Cost Center negative tests', () => {
...
-        })
-      );
+
+      for (const invalidRequest of invalidCreateRequests) {
+        try {
+          await itAdminSession.resources.costCenters.create(invalidRequest, false);
+        } catch (error) {
+          checkHttpError(
+            error,
+            new HttpError(400, {
```
SanketD92, line 148: same comment as above

PR #1159, opened 2023-06-21
feat: cost center account status error

aws-tyler:
Description of changes:
Throw a 400 if the account being used to create the cost center is not in the `CURRENT` state.

### DIFF
--- a/solutions/swb-app/src/costCenterRoutes.ts
+++ b/solutions/swb-app/src/costCenterRoutes.ts
@@ -13,7 +13,8 @@ import {
   UpdateCostCenterRequestParser,
   DeleteCostCenterRequest,
   DeleteCostCenterRequestParser,
-  ProjectService
+  ProjectService,
+  isInvalidAccountStateError
 } from '@aws/workbench-core-accounts';
 import { isInvalidPaginationTokenError, validateAndParse } from '@aws/workbench-core-base';
 import * as Boom from '@hapi/boom';
@@ -33,7 +34,16 @@ export function setUpCostCenterRoutes(
         CreateCostCenterRequestParser,
         req.body
       );
-      res.status(201).send(await costCenterService.create(validatedRequest));
+
+      try {
+        res.status(201).send(await costCenterService.create(validatedRequest));
+      } catch (e) {
+        if (isInvalidAccountStateError(e)) {
+          Boom.badRequest(e.message);
+        }
+
+        throw e;
+      }
     })
   );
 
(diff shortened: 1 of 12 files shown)


workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -131,7 +132,20 @@ export default class CostCenterService {
...
     const id = uuidWithLowercasePrefix(resourceTypeToKey.costCenter);
-    const account = await this._getAccount(request.accountId);
+    let account: Account;
+
+    const accountService = new AccountService(this._dynamoDbService);
+    try {
+      account = await accountService.getAccount(request.accountId);
+    } catch (e) {
+      console.error(`Failed to get account for cost center creation: ${e.message}`);
+      throw Boom.badRequest(`Failed to get account for cost center creation ${request.accountId}`);
```
kpark277, line 142: Per app sec guidelines, do not reflect user inputs on error.

solutions/swb-app/src/costCenterRoutes.ts
```diff
@@ -33,7 +34,16 @@ export function setUpCostCenterRoutes(
         CreateCostCenterRequestParser,
         req.body
       );
-      res.status(201).send(await costCenterService.create(validatedRequest));
+
+      try {
```
maghirardelli, line 38: nit: should catch all errors, check `isBoomError` and finally throw a `Boom.badImplementation` error if it is not a known error type

workbench-core/accounts/src/errors/errors.test.ts
```diff
@@ -11,4 +12,10 @@ describe('custom error tests', () => {
 
     expect(isInvalidAwsAccountIdError(invalidAwsAccountIdError)).toBe(true);
   });
+
+  test('InvalidAccountError', () => {
+    const error = new InvalidAccountStateError();
+
+    expect(isInvalidAccountStateError(error)).toBe(true);
```
maghirardelli, line 19: nit: should also check that isValidAccountStateError can evaluate to false if the passed error is of a different type

workbench-core/accounts/src/services/costCenterService.ts
```diff
@@ -131,7 +132,20 @@ export default class CostCenterService {
 
   public async create(request: CreateCostCenterRequest): Promise<CostCenter> {
     const id = uuidWithLowercasePrefix(resourceTypeToKey.costCenter);
-    const account = await this._getAccount(request.accountId);
+    let account: Account;
+
+    const accountService = new AccountService(this._dynamoDbService);
```
maghirardelli, line 137: Why take this out of the helper method?

maghirardelli (approved):

workbench-core/accounts/src/services/costCenterService.ts
aws-tyler: the helper method was only a couple lines long and hid another throw that could happen. Since it doesn't need to be shared, I felt it would make the code easier to reason about because you can see what failure scenarios you need to handle.

insignias (approved):

